PR #6, opened 2022-10-03
Add paths for tree graphs

leburgel:
Request to add some functionalities for finding paths on tree graphs, which are particularly useful for certain tree tensor network routines. A summary of the changes:

- Added a function `vertex_path(graph, s, t)` which returns the list of vertices along the path from the source vertex `s` to the target vertex `t`. For a directed tree, if there is no path between `s` and `t` the function returns `nothing`.
- Added a function `edge_path(graph, s, t)` which returns the list of edges along the path from the source vertex `s` to the target vertex `t` (where the edges are directed from `s` to `t`). For a directed tree, if there is no path between `s` and `t` the function returns `nothing`.
- Small change to `parent_vertex` such that it returns `nothing` when requesting the parent vertex of the tree root (as this case would throw an error before).
- Change in the definition used for a leaf vertex of a rooted directed tree to a vertex with no children, as well as fixing some typos in related methods.

### DIFF
--- a/src/Graphs/abstractgraph.jl
+++ b/src/Graphs/abstractgraph.jl
@@ -43,6 +43,19 @@ function incident_edges(graph::AbstractGraph, vertex...)
   ]
 end
 
+# Get the leaf vertices of a tree-like graph
+# 
+# For the directed case, could also use `AbstractTrees`:
+#
+# root_index = findfirst(vertex -> length(outneighbors(vertex)) == length(neighbors(vertex)), vertices(graph))
+# root = vertices(graph)[root_index]
+# [node.vertex for node in Leaves(TreeGraph(graph, root))]
+#
+function leaf_vertices(graph::AbstractGraph)
+  # @assert is_tree(graph)
+  return filter(v -> is_leaf(graph, v...), vertices(graph))
+end
+
 #
 # Graph iteration
 #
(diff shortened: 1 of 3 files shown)


src/Graphs/abstractgraph.jl
```diff
@@ -121,7 +136,7 @@ end
 #
 @traitfn function leaf_vertices(graph::::IsDirected)
   # @assert is_tree(graph)
-  return [vertex for vertex in vertices(graph) if isone(length(inneighbors(vertex)))]
+  return [vertex for vertex in vertices(graph) if isempty(outneighbors(graph, vertex))]
```
mtfishman, line 139: I thought I already commented this but I can't find it now.

Can we use `is_leaf(graph, vertex)` here?

leburgel: Sure, I guess we could just catch both the undirected and directed versions as
```Julia
function leaf_vertices(graph::AbstractGraph)
  return filter(v -> is_leaf(graph, v...), vertices(graph))
end
```
and leave the rest up to `leaf_vertex`.

mtfishman: Do you mean `is_leaf`? That sounds like a good plan.

leburgel: Yes I do, sorry about the typo.

mtfishman:
Thanks @leburgel!

It looks like for undirected trees you make a full directed tree first, and then use the algorithm for directed trees. I wonder if there are faster ways to do that. Maybe even a general shortest path algorithm ([for example](https://juliagraphs.org/Graphs.jl/dev/algorithms/shortestpaths/#Graphs.a_star-Union{Tuple{E},%20Tuple{U},%20Tuple{T},%20Tuple{AbstractGraph{U},%20Integer,%20Integer},%20Tuple{AbstractGraph{U},%20Integer,%20Integer,%20AbstractMatrix{T}},%20Tuple{AbstractGraph{U},%20Integer,%20Integer,%20AbstractMatrix{T},%20Function},%20Tuple{AbstractGraph{U},%20Integer,%20Integer,%20AbstractMatrix{T},%20Function,%20Type{E}}}%20where%20{T,%20U,%20E%3C:AbstractEdge})) would work well? Anyway, that seems like something we could explore going forward, but it is a nice feature that will be very useful to have so we can leave optimization for future work.

PR #7, opened 2022-10-03
Extend comb tree constructors

leburgel:
Request to extend the constructors for comb trees to handle cases where not all 'teeth' of the comb have the same length. I've added a method which takes a vector of integers specifying the length of each tooth. The vertex labeling remains the same as that of a 2d grid, where now some vertices of the underlying grid can be missing.


mtfishman:
Looks good, thanks!

PR #8, opened 2022-10-03
Add some AbstractNamedGraph functionalities

leburgel:
Some convenience `AbstractNamedGraph` functionalities that I ended up using quite often, so I though it could make sense to add them here. A summary of the (potential) additions:
- A function `rename_vertices` which allows renaming the vertices of an `AbstractNamedGraph` or `AbstractNamedEdge` given a `Function` or `Dictionary` name map. It is basically the same as the one in https://github.com/mtfishman/ITensorNetworks.jl/blob/main/examples/peps/utils.jl, with some minor adjustments to make it work for all currently existing `AbstractNamedGraph` subtypes. In particular, it seemed to make sense to allow renaming to change the vertex type for the case of `NamedGraph`.
- Constructors for all currently existing `AbstractNamedGraph` subtypes which just take a list of vertices, similar to how `Graphs.Graph(::Integer)` returns a graph with a given number of vertices and no edges.
- Equality for `AbstractNamedGraph`s, useful for checking if the underlying graph of two instances of a given data structure is the same.

### DIFF
--- a/src/abstractnamededge.jl
+++ b/src/abstractnamededge.jl
@@ -30,6 +30,6 @@ set_dst(e::AbstractNamedEdge, dst) = set_vertices(e, src(e), dst)
 
 rename_vertices(f::Function, e::AbstractNamedEdge) = set_vertices(e, f(src(e)), f(dst(e)))
 
-function rename_vertices(e::AbstractNamedEdge, name_map::Dictionary)
+function rename_vertices(e::AbstractNamedEdge, name_map)
   return rename_vertices(v -> name_map[v], e)
 end
--- a/src/abstractnamededge.jl
+++ b/src/abstractnamededge.jl
@@ -24,3 +24,9 @@ function ==(e1::AbstractNamedEdge, e2::AbstractNamedEdge)
   return (src(e1) == src(e2) && dst(e1) == dst(e2))
 end
 hash(e::AbstractNamedEdge, h::UInt) = hash(src(e), hash(dst(e), h))
+
+function rename_vertices(e::ET, name_map::Dictionary) where {ET<:AbstractNamedEdge}
+  # strip type parameter to allow renaming to change the vertex type
+  base_edge_type = Base.typename(ET).wrapper
+  return base_edge_type(name_map[src(e)], name_map[dst(e)])
+end
--- a/src/abstractnamededge.jl
+++ b/src/abstractnamededge.jl
@@ -25,8 +25,11 @@ function ==(e1::AbstractNamedEdge, e2::AbstractNamedEdge)
 end
 hash(e::AbstractNamedEdge, h::UInt) = hash(src(e), hash(dst(e), h))
 
-function rename_vertices(e::ET, name_map::Dictionary) where {ET<:AbstractNamedEdge}
-  # strip type parameter to allow renaming to change the vertex type
-  base_edge_type = Base.typename(ET).wrapper
-  return base_edge_type(name_map[src(e)], name_map[dst(e)])
+set_src(e::AbstractNamedEdge, src) = set_vertices(e, src, dst(e))
+set_dst(e::AbstractNamedEdge, dst) = set_vertices(e, src(e), dst)
+
+rename_vertices(f::Function, e::AbstractNamedEdge) = set_vertices(e, f(src(e)), f(dst(e)))
+
+function rename_vertices(e::AbstractNamedEdge, name_map::Dictionary)
+  return rename_vertices(v -> name_map[v], e)
 end
--- a/src/abstractnamedgraph.jl
+++ b/src/abstractnamedgraph.jl
@@ -237,3 +237,33 @@ function show(io::IO, mime::MIME"text/plain", graph::AbstractNamedGraph)
 end
 
 show(io::IO, graph::AbstractNamedGraph) = show(io, MIME"text/plain"(), graph)
+
+# 
+# Convenience functions
+#
+
+function Base.:(==)(g1::GT, g2::GT) where {GT<:AbstractNamedGraph}
+  issetequal(vertices(g1), vertices(g2)) || return false
+  for v in vertices(g1)
+    issetequal(inneighbors(g1, v), inneighbors(g2, v)) || return false
+    issetequal(outneighbors(g1, v), outneighbors(g2, v)) || return false
+  end
+  return true
+end
+
+function rename_vertices(g::GT, name_map::Dictionary) where {GT<:AbstractNamedGraph}
+  original_vertices = vertices(g)
+  new_vertices = getindices(name_map, original_vertices)
+  # strip type parameter to allow renaming to change the vertex type
+  base_graph_type = Base.typename(GT).wrapper
+  new_g = base_graph_type(new_vertices)
+  for e in edges(g)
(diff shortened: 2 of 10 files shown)


src/abstractnamededge.jl
```diff
@@ -24,3 +24,9 @@ function ==(e1::AbstractNamedEdge, e2::AbstractNamedEdge)
   return (src(e1) == src(e2) && dst(e1) == dst(e2))
 end
 hash(e::AbstractNamedEdge, h::UInt) = hash(src(e), hash(dst(e), h))
+
+function rename_vertices(e::ET, name_map::Dictionary) where {ET<:AbstractNamedEdge}
+  # strip type parameter to allow renaming to change the vertex type
+  base_edge_type = Base.typename(ET).wrapper
```
mtfishman, line 30: This is a tricky design choice here.

I think it is best to avoid using internal Julia features like this if possible. An alternaive design I can think of is that for each `AbstractNamedEdge` we should expect them to define generic functions `set_src` and `set_dst`, for example:
```julia
set_src(e::NamedEdge, src) = NamedEdge(src, dst(e))
set_dst(e::NamedEdge, dst) = NamedEdge(src(e), dst)
```
Then the concrete types can decide how to infer the types of the resulting edges based on the new vertex types.

Another thing that comes to mind is that we can have a more generic `map`-like interface:
```julia
function rename_vertices(f::Function, e::AbstractNamedEdge)
  #...
end
```
where the `Dictionary` interface is just a special case:
```julia
function rename_vertices(e::AbstractNamedEdge, name_map)
  return rename_vertices(v -> name_map[v], e)
end
```
This could be a way to do simple maps like appending a dimension to a vertex, for example:
```julia
rename_vertices(v -> (v..., 1), e)
```

src/abstractnamedgraph.jl
```diff
@@ -237,3 +237,33 @@ function show(io::IO, mime::MIME"text/plain", graph::AbstractNamedGraph)
...
+    issetequal(outneighbors(g1, v), outneighbors(g2, v)) || return false
+  end
+  return true
+end
+
+function rename_vertices(g::GT, name_map::Dictionary) where {GT<:AbstractNamedGraph}
+  original_vertices = vertices(g)
+  new_vertices = getindices(name_map, original_vertices)
+  # strip type parameter to allow renaming to change the vertex type
+  base_graph_type = Base.typename(GT).wrapper
```
mtfishman, line 258: Similar comment as above. We should try to use a generic function like `set_vertices` that would be defined for every `AbstractNamedGraph`.

src/abstractnamedgraph.jl
```diff
@@ -237,3 +237,33 @@ function show(io::IO, mime::MIME"text/plain", graph::AbstractNamedGraph)
...
+  new_g = base_graph_type(new_vertices)
+  for e in edges(g)
+    add_edge!(new_g, rename_vertices(e, name_map))
+  end
+  return new_g
+end
+
+function rename_vertices(g::AbstractNamedGraph, name_map::Function)
```
mtfishman, line 266: Nice, I see you already used my suggestion of having a function interface. I think it would be best to have this structured the opposite way, where the function version is the internal version and the `Dictionary` version is just a special case. Also, I would put `name_map` first here, since it is conventional in Julia to have functions go first in map-like higher-order functions so that you can use the `do`-block syntax (https://docs.julialang.org/en/v1/manual/functions/#Do-Block-Syntax-for-Function-Arguments).

src/abstractnamededge.jl
leburgel: I agree I should have stayed out of Julia internals. `set_src`, `set_dst` and the `map`-like interface for `rename_vertices` all sound like good suggestions.

One issue that comes to mind is that `rename_vertices(f::Function, e::AbstractNamedEdge)` can't just do `set_src(e, f(src(e))` and `set_src(e, f(dst(e))` sequentially in general, because renaming one vertex at a time can lead to incompatible vertex types (e.g. when renaming integer vertex labels to string vertex labels for a `NamedEdge`).

We could just define a specific
```Julia
rename_vertices(f::Function, e::NamedEdge) = NamedEdge(f(src(e)), f(dst(e)))
```
for every `AbstractNamedEdge` subtype in addition to `set_src`, `set_dst`, which again lets the concrete types decide how to infer the types of the resulting edges based on the new vertex types while allowing that the vertex types to remain compatible during renaming.

src/abstractnamedgraph.jl
leburgel: Sounds good. Do you imagine `set_vertices` to take a list of vertex names as input, or should it also work with a general name map function (in which case it would in essence be a `rename_vertices` for `AbstractNamedGraph`s)?

mtfishman: I think it should be a lower level and internal function that just takes a Vector of vertex names (and therefore relies on knowing the internal ordering of the vertices of the parent graph).

src/abstractnamededge.jl
mtfishman: Good point about the types. I was picturing that the `NamedEdge`/`NamedDimEdge` typing was more flexible and could have mixtures of vertex types but I guess that's not the case. Probably we should expand the definitions to:
```julia
struct NamedEdge{S,D} <: AbstractNamedEdge{S,D}
  src::S
  dst::D
end
```
since in general we'll want graphs that have mixtures of vertex types but we don't need to handle that in this PR.

We could define a generic function `set_vertices(e::AbstractNamedEdge, src, dst)`:
```julia
set_vertices(e::NamedEdge, src, dst) = NamedEdge(src, dst)
```
and then define:
```julia
rename_vertices(f::Function, e::AbstractNamedEdge) = set_vertices(e, f(src(e)), f(dst(e)))
```
to avoid having to define `rename_vertices` for each `AbstractNamedEdge`.

leburgel:
As per the suggestions above, I switched to a core `map`-like interface for `rename_vertices` for `AbstractNamedEdge` and `AbstractNamedGraph`, with a `Dictionary` name map as a special case. These now rely on an internal `set_vertices` routine defined for each concrete subtype.

src/abstractnamededge.jl
```diff
@@ -24,3 +24,12 @@ function ==(e1::AbstractNamedEdge, e2::AbstractNamedEdge)
   return (src(e1) == src(e2) && dst(e1) == dst(e2))
 end
 hash(e::AbstractNamedEdge, h::UInt) = hash(src(e), hash(dst(e), h))
+
+set_src(e::AbstractNamedEdge, src) = set_vertices(e, src, dst(e))
+set_dst(e::AbstractNamedEdge, dst) = set_vertices(e, src(e), dst)
+
+rename_vertices(f::Function, e::AbstractNamedEdge) = set_vertices(e, f(src(e)), f(dst(e)))
+
+function rename_vertices(e::AbstractNamedEdge, name_map::Dictionary)
```
mtfishman, line 33: We could just leave the type off here to allow users to input any dictionary-like object they want (or anything that implements `getindex`).

leburgel: Done.

mtfishman:
Looks great, thanks a lot! Also thanks for adding all those tests, these packages have been undertested...

PR #13, opened 2022-11-28
Remove `NamedDimGraph`, refactor and simplify

mtfishman:
This removes the `NamedDimGraph` type, along with slicing syntax `g[:, 1]` in favor of the more verbose but general `subgraph(v -> v[2] == 1, g)` syntax. The slicing syntax might return, but it causes some syntax conflicts with `DataGraphs` metadata access through `getindex` so for simplicity I am removing it for now.


mtfishman:
@leburgel @JoeyT1994 heads up about some interface changes. I'm updating DataGraphs.jl and ITensorNetworks.jl accordingly. Over all this should simplify things (like adding support for Graphs.jl functionality, not having too many graph types floating around, etc.) and over all shouldn't change high level ITensorNetworks functionality.

PR #29, opened 2023-02-02
Adding and Removing Edge Lists

JoeyT1994:
Added utility functions for adding and removing lists of edges.


mtfishman:
Also could you add some tests?

JoeyT1994:
> Also could you add some tests?
Sure, I just added them.

test/test_add_rem_edges.jl
```diff
@@ -0,0 +1,18 @@
...
+using NamedGraphs
+using NamedGraphs: add_edges, add_edges!, rem_edges, rem_edges!
+using Graphs
+
+
+@testset "Adding and Removing Edge Lists" begin
+
+    g = named_grid((2, 2))
+    rem_edges!(g, [(1,1) => (1,2), (1,1) => (2,1)])
+    @test !has_edge(g, (1,1) => (1,2)) && !has_edge(g, (1,1) => (2,1)) && has_edge(g, (1,2) => (2,2))
```
mtfishman, line 11: Better to put them on separate lines:
```julia
@test !has_edge(g, (1,1) => (1,2))
@test !has_edge(g, (1,1) => (2,1))
# etc.
```
so then you can see which one fails, if only one fails.

mtfishman:
I added some checks for formatting. Could you format your code:
```julia
julia> cd(pkgdir(NamedGraphs))

julia> using JuliaFormatter

julia> format(".")
```

src/Graphs/abstractgraph.jl
```diff
@@ -379,3 +379,35 @@ function mincut_partitions(graph::AbstractGraph, distmx=weights(graph))
...
+
+"""Add a list of edges to a graph g"""
+function add_edges(g::AbstractGraph, edges)
+  g_out = copy(g)
+  for e in edges
+    add_edge!(g_out, e)
+  end
+
+  return g_out
+end
```
mtfishman, line 407: I think it would be better to implement in terms of `add_edges!`, i.e.:
```julia
function add_edges(g::AbstractGraph, edges)
  g = copy(g)
  add_edges!(g, edges)
  return g
```
and same for `rem_edges`. Best to not duplicate code if you don't need to.

src/Graphs/abstractgraph.jl
```diff
@@ -379,3 +379,35 @@ function mincut_partitions(graph::AbstractGraph, distmx=weights(graph))
...
+
+function add_edges!(g::AbstractGraph, edges)
+  for e in edges
+    add_edge!(g, e)
+  end
```
mtfishman, line 412: I would `return g` from this function, as a convention I like to return the object that I modify so that someone could write `g = add_edges!(g, es)`. Same for `rem_edges!`.

src/Graphs/abstractgraph.jl
```diff
@@ -381,33 +381,32 @@ function mincut_partitions(graph::AbstractGraph, distmx=weights(graph))
...
 
-"""Add a list of edges to a graph g"""
-function add_edges(g::AbstractGraph, edges)
+function rem_edges(g::AbstractGraph, edges)
   g_out = copy(g)
-  for e in edges
-    add_edge!(g_out, e)
-  end
-
+  rem_edges!(g, edges)
```
mtfishman, line 393: Should be `g_out`, or rename `g_out` to `g` (that would be my preference).

mtfishman:
Thanks!

PR #32, opened 2023-02-13
Random bfs tree

JoeyT1994:
This PR adds a function (and testing) for constructing a Breadth First Search tree for a graph but in a random manner.

The tree stems from a source vertex `s`. The edges from `s` to its neighbors are guaranteed to be in the tree but remaining edges in the tree will be random.

### DIFF
--- a/src/Graphs/abstractgraph.jl
+++ b/src/Graphs/abstractgraph.jl
@@ -411,3 +411,32 @@ function add_edges(g::AbstractGraph, edges)
   add_edges!(g, edges)
   return g
 end
+
+""" Do a BFS search to construct a tree, but do it with randomness to avoid generating the same tree """
+function random_bfs_tree(g::AbstractGraph, s)
+  Q = [s]
+  d = Dict([v ≠ s ? (v, Inf) : (v, 0) for v in vertices(g)])
+  visited = [s]
+  g_out = NamedDiGraph(vertices(g))
+
+  while !isempty(Q)
+    v = rand(Q)
+    setdiff!(Q, [v])
+    for vn in neighbors(g, v)
+      if(d[vn] > d[v] + 1)
+        d[vn] = d[v] + 1
+        if(vn ∉ Q)
+          if(vn ∉ visited)
+            add_edge!(g_out, v => vn)
+            push!(visited, vn)
+          end
+          push!(Q, vn)
+        end
+      end
+    end
+  end
+
+  return g_out
+
+
+end
--- a/src/Graphs/abstractgraph.jl
+++ b/src/Graphs/abstractgraph.jl
@@ -412,7 +412,7 @@ function add_edges(g::AbstractGraph, edges)
   return g
 end
 
-""" Do a BFS search to construct a tree, but do it with randomness to avoid generating the same tree """
+""" Do a BFS search to construct a tree, but do it with randomness to avoid generating the same tree. Based on Int. J. Comput. Their Appl. 15 pp 177-186 (2008). Edges will point away from source vertex s."""
 function random_bfs_tree(g::AbstractGraph, s)
   Q = [s]
   d = Dict([v ≠ s ? (v, Inf) : (v, 0) for v in vertices(g)])
@@ -437,6 +437,5 @@ function random_bfs_tree(g::AbstractGraph, s)
   end
 
   return g_out
-
 
 end
--- a/test/test_random_bfs_tree.jl
+++ b/test/test_random_bfs_tree.jl
@@ -1,0 +1,24 @@
+using Test
+using NamedGraphs
+using NamedGraphs: random_bfs_tree
+using Graphs
+using Random
+
+@testset "Random BFs Tree" begin
+  g = named_grid((10, 10))
+
+  s = (5,5)
+
+  Random.seed!(1234)
+  g_randtree1 = random_bfs_tree(g, s)
+  g_nonrandtree1 = bfs_tree(g, s)
+  Random.seed!(1434)
+  g_randtree2 = random_bfs_tree(g, s)
+  g_nonrandtree2 = bfs_tree(g, s)
+
+  @test length(edges(g_randtree1)) == length(vertices(g_randtree1)) - 1 && is_connected(g_randtree1)
+  @test length(edges(g_randtree2)) == length(vertices(g_randtree2)) - 1 && is_connected(g_randtree2)
+
+  @test edges(g_randtree1) != edges(g_randtree2)
+  @test edges(g_nonrandtree1) == edges(g_nonrandtree2)
+end
--- a/test/test_random_bfs_tree.jl
+++ b/test/test_random_bfs_tree.jl
@@ -4,7 +4,7 @@ using NamedGraphs: random_bfs_tree
 using Graphs
 using Random
 
-@testset "Random BFs Tree" begin
+@testset "Random Bfs Tree" begin
   g = named_grid((10, 10))
 
   s = (5,5)


src/Graphs/abstractgraph.jl
```diff
@@ -411,3 +411,31 @@ function add_edges(g::AbstractGraph, edges)
   add_edges!(g, edges)
   return g
 end
+
+""" Do a BFS search to construct a tree, but do it with randomness to avoid generating the same tree. Based on Int. J. Comput. Their Appl. 15 pp 177-186 (2008). Edges will point away from source vertex s."""
+function random_bfs_tree(g::AbstractGraph, s)
+  Q = [s]
+  d = Dict([v ≠ s ? (v, Inf) : (v, 0) for v in vertices(g)])
+  visited = [s]
+  g_out = NamedDiGraph(vertices(g))
```
mtfishman, line 420: Maybe we could use `directed_graph(typeof(g))(vertices(g))` to try to preserve the type of the input graph. For example, it would be nice if it returns a `SimpleDiGraph` in the case where the input graph is a `SimpleGraph`.

Doing that kind of generic programming can be tough in general so it isn't critical if it doesn't seem easy to get working, we could always improve on that later.

JoeyT1994: Yeah that would be ideal but I think the issue with that would be that `SimpleDiGraph(Vector{Int64})` (which is what would get called if you gave `g` as a simple graph) isn't defined and it needs to be `SimpleDiGraph(length(Vector{Int64}))` or something, which is different from the `NamedGraph` case.

Obviously could check for such a case but it seems a bit messy. What do you think?

mtfishman: Yeah I thought that might be a challenge since they have different constructor interfaces (making a `SimpleDiGraph` without any edges expects an integer representing the number vertices, while a `NamedGraph` generally needs a list of vertices).

It seems like the correct concept in Julia would be `similar`, for example for sparse arrays:
```julia
julia> using SparseArrays

julia> A = sprandn(4, 4, 0.5)
4×4 SparseMatrixCSC{Float64, Int64} with 7 stored entries:
   ⋅         ⋅     ⋅          ⋅ 
 -0.21712    ⋅   -0.741368   0.199437
   ⋅         ⋅     ⋅         1.78635
  0.486078   ⋅    0.593271  -0.199654

julia> similar(A, 4, 4)
4×4 SparseMatrixCSC{Float64, Int64} with 0 stored entries:
  ⋅    ⋅    ⋅    ⋅ 
  ⋅    ⋅    ⋅    ⋅ 
  ⋅    ⋅    ⋅    ⋅ 
  ⋅    ⋅    ⋅    ⋅ 
```

Here we could define:
```julia
Base.similar(G::Type{<:AbstractGraph}, vertices)
Base.similar(G::Type{<:AbstractGraph}, nvertices::Integer)
```
as a way to generically create a graph of the same type `G` without any edges with either the specified vertices or number of vertices. Then we could use:
```julia
dg = similar(directed_graph(typeof(g)), vertices(g))
```
We could leave this for a future issue since that is a more involved thing that will require some thought about how to do it generically.

JoeyT1994: Okay nice, that seems like something worth doing in the future to help bridge the different constructors for `NamedGraphs` and `SimpleGraphs`

src/Graphs/abstractgraph.jl
```diff
@@ -411,3 +411,30 @@ function add_edges(g::AbstractGraph, edges)
   add_edges!(g, edges)
   return g
 end
+
+""" Do a BFS search to construct a tree, but do it with randomness to avoid generating the same tree. Based on Int. J. Comput. Their Appl. 15 pp 177-186 (2008). Edges will point away from source vertex s."""
+function random_bfs_tree(g::AbstractGraph, s)
+  Q = [s]
+  d = Dict([v ≠ s ? (v, Inf) : (v, 0) for v in vertices(g)])
```
mtfishman, line 418: Maybe:
```julia
d = map(v -> v == s ? 0.0 : Inf, Indices(vertices(g)))
```
That is syntax from `Dictionaries.jl`, I personally like the `Dictionaries.Dictionary` interface better than `Base.Dict`. I kept rereading the current version and wondering why you were making a list of tuples until I realized that the `Dict` constructor interprets them as key-value pairs.

src/Graphs/abstractgraph.jl
```diff
@@ -411,3 +411,30 @@ function add_edges(g::AbstractGraph, edges)
...
+  visited = [s]
+  g_out = NamedDiGraph(vertices(g))
+
+  while !isempty(Q)
```
mtfishman, line 422: I generally feel uncomfortable using `while` loops like this since it is just asking for infinite loops.

Can we come up with an upper bound on the number of times this could be expected to iterate and add a check for that to make sure it stops at some point?

mtfishman: We can then add a check afterward that it completed properly, i.e. check that `isempty(Q)` after the loop and throw an error if not.

mtfishman:
```julia
isempty_Q = false
for iter in 1:max_iter

  # Your code here

  isempty_Q = isempty(Q)
  if isempty_Q
    break
  end
end
if !isempty_Q
  error("Error message that this function failed")
end
```

JoeyT1994: Good idea. The required number of iterations will scale as O(vertices(g) + edges(g)) but I am not sure how to come up with an upper bound other than that something like c*(vertices(g)+edges(g))^{2} should be safe provided c is large enough (although it is not immediate what that should be - something like a 1000 is probably very safe).

mtfishman: Maybe just `c * (nv(g) + ne(g))`? `c *  (nv(g) + ne(g))^2` can get very large:
```julia
julia> log10(1000 * (100 + 100^2)^2)
11.008642747565284

julia> log10(1000 * (100 + 100^2))
7.004321373782642
```
Of course checking `!isempty_Q` will tell us if it didn't complete correctly (or checking it isn't actually a spanning tree, but maybe that's the same thing?) so we can try out something a bit stricter and see how it goes.

src/Graphs/abstractgraph.jl
```diff
@@ -413,20 +413,19 @@ function add_edges(g::AbstractGraph, edges)
 end
 
 """ Do a BFS search to construct a tree, but do it with randomness to avoid generating the same tree. Based on Int. J. Comput. Their Appl. 15 pp 177-186 (2008). Edges will point away from source vertex s."""
-function random_bfs_tree(g::AbstractGraph, s; max_iter = nothing)
+function random_bfs_tree(g::AbstractGraph, s; max_iter=nothing)
```
mtfishman, line 416: I would write this as:
```julia
function random_bfs_tree(g::AbstractGraph, s; maxiter=(1000 * (nv(g) + ne(g)^2)))
```
(renaming `max_iter` to `maxiter` and dropping the if-statement below). To clarify, does the upper bound scale as `O(nv(g) + ne(g)^2)` or `O(nv(g) + ne(g))`?

JoeyT1994: Okay good, I didn't realise you could refer to arguments of the function within keyword argument defaults.
Expected number of iters scales as O(nv(g) + ne(g)). I imagine that for a given finite epsilon << 1 you can prove there is some C that the algorithm is, with probability 1- epsilon, guaranteed to finish within C * (nv(g) + ne(g)) iterations. Assumedly C = 1000 (have tested for some quite big lattices) is enough to cover epsilon very small but I can't be more precise than that.

mtfishman: Ok, I'm just getting confused between using `1000 * (nv(g) + ne(g)^2)` vs. `1000 * (nv(g) + ne(g))` since I've seen different ones mentioned in the discussion. So shouldn't we use `1000 * (nv(g) + ne(g))` as the default since that's the theoretical scaling? It seems like in the latest you used `1000 * (nv(g) + ne(g)^2)`.

JoeyT1994: Sorry yes that is a mistake, I meant to put `1000 * (nv(g) + ne(g))` in. Have fixed it now.

src/Graphs/abstractgraph.jl
```diff
@@ -411,3 +411,39 @@ function add_edges(g::AbstractGraph, edges)
...
+      end
+    end
+
+    isempty_Q = isempty(Q)
+    if isempty_Q
+      break
+    end
+  end
+  if !isempty_Q
+    error("Search failed to cover the graph in time. Consider increasing max_iter.")
```
mtfishman, line 445: Should be `maxiter`.

mtfishman:
Thanks!

Issue #36, opened 2023-07-11
Constructors for decorated graphs


JoeyT1994:
We should think about having construction methods for decorated graphs.

E.g. given a named graph `g` we have some function like

`decorate_graph(g::NamedGraph, edge_map::Function)`

where `edge_map::Function` is a map which takes the edges of `g` and outputs the namedgraph which will then be inserted along the edges of the original graph `g`,

For example something like 
```
g = named_grid((L,L))
g_dec = decorate_graph(g::NamedGraph, e -> named_grid((1)))
```

would build the `L x L` Lieb lattice by inserting a single vertex along the edges of the regular square lattice.

Obviously we would have to think about how to name the new vertices that have been added to `g` and also how to specify  which vertices of the inserted graph connect to `src(e)` and `dst(e)` (perhaps by defaulting to the first vertex of the decorated graph with additional specification if desired).

mtfishman:
We could have a naming convention like `(e, v)` where `e` is the edge where the edge decoration graph is being inserted and `v` is the name of the vertex of the edge decoration graph. Or it is just up to the user to specify names based on the edge input, for example:
```julia
g = named_grid((L,L))
function edge_decoration_graph(e)
  g_e = named_graph()
  add_vertex!(g_e, src(e))
  add_vertex!(g_e, (e, 1))
  add_vertex!(g_e, dst(e))
  add_edge!(g_e, src(e), (e, 1))
  add_edge!(g_e, (e, 1), dst(e))
  return g_e
end
g_dec = decorate_graph(g::NamedGraph; edge_decoration=edge_decoration_graph)
```
but the convention of automatically adding an edge from `src(e)` to the first vertex and an edge from the last vertex to `dst(e)` could make sense, obviously that makes defining the edge decoration graphs simpler.

Also a complementary functionality would be to replace the vertices of a graph with a vertex decoration graph, for example:
```julia
function vertex_decoration_graph(v)
  g_v = named_graph()
  # ...
end
g_dec = decorate_graph(g::NamedGraph; vertex_decoration=vertex_decoration_graph)
```

JoeyT1994:
I think that naming convention makes the most sense and we should have that as default. Also having vertex replacement makes sense too

JoeyT1994:
I just realised that with this we would create the 127-qubit IBM heavy hex lattice as follows:
```
g = hexagonal_lattice_graph(3, 5)
[...]

PR #37, opened 2023-07-14
Decorated graphs

JoeyT1994:
This PR, which addresses #36, adds functionality for decorating graphs by adding additional graphs on their edges or vertices.

Two functions are provided 

`decorate_graph_edges(g::NamedGraph; edge_map::Function=e -> named_grid((1,)))`

and 

`decorate_graph_vertices(g::NamedGraph; vertex_map::Function=v -> named_grid((1,)))`

Currently there are defaults for which vertices of the decoration are connected to the edges/vertices original graph but we will add keyword arguments for this in the future.

Testing is included. This PR also fixed an ambiguity issue with function `a_star` when passed a `NamedGraph` with Integer labelling.


src/abstractnamedgraph.jl
```diff
@@ -295,6 +295,29 @@ function a_star(
   return parent_edges_to_edges(graph, parent_shortest_path)
 end
 
+function a_star(
+  graph::AbstractNamedGraph,
+  source,
+  destination,
+  distmx=weights(graph),
+  heuristic::Function=(v -> zero(eltype(distmx))),
+  edgetype_to_return=edgetype(graph),
```
mtfishman, line 304: We could use `args...` here so we don't have to list the arguments explicitly (besides the ones involved in the ambiguity error). Same for the more narrow `a_star` below.

mtfishman: Also, could you add a short test for `a_star` with a named graph with integer vertices?

mtfishman:
For the current version of `vertex_map`, aren't the edges connecting to the decorated vertex essentially random since the ordering that `neighbors` outputs is an internal detail? Seems like until we decide how to specify those new edges, it's not so useful.

mtfishman:
Maybe `vertex_map` could be a function that accepts a list of vertices, where the first one is the vertex and the rest are the neighbors of that vertex. Then the user is expected to add edges between their new vertices in the vertex decoration and the neighbor vertices. If those aren't added, then those edges could be added (essentially) randomly as you do it now.

src/abstractnamedgraph.jl
JoeyT1994: Sure I will do that

JoeyT1994:
> For the current version of `vertex_map`, aren't the edges connecting to the decorated vertex essentially random since the ordering that `neighbors` outputs is an internal detail? Seems like until we decide how to specify those new edges, it's not so useful.

Currently the vertex `vd` which is being decorated is removed and replaced with the decoration graph `gd.` The edges which originally connected `vd` to its neighbors in the undecorated graph `vd_neighbors` are reconnected to the first vertex of `gd`? So there is nothing random going on but obviously ideally the user should decide which vertices of `gd` should connect up to which members of `vd_neighbors` ?

mtfishman:
Oh I see, I was picturing that each vertex neighbor would get connected to unique vertices of the decorated graph getting inserted onto the vertex, that's why I thought it would be random to some extent. In that case, I agree it isn't random since the user can choose which is the first vertex of the decorated vertex graph. In that case it seems fine to leave it for future work to give more control to the users over how the edges to the neighbors get specified.

PR #39, opened 2023-10-19
Added forest cover function

JoeyT1994:
This PR adds functionality for creating a forest cover of a graph `g` (see [wiki page](https://en.wikipedia.org/wiki/Arboricity) ). Specifically, a forest cover is a set of forests which collectively cover all the edges in `g`. In our case, each forest in the forest cover covers all vertices in `g` (although this is not strictly necessary).

Testing is included and the number of forests appears fairly lightweight. Whilst our algorithm does not construct the minimum number of forests it appears competitive and reasonable.


JoeyT1994:
Okay I am confused. These tests are all passing on my local machine? Including testing on Ubuntu w/ Julia 1.7.2?

mtfishman:
Do the tests rely on some randomness? That can change between Julia versions.

mtfishman:
Does this cover edges in both directions? Or would that be handled through the input, i.e. inputting a directed graph that has both edge directions?

mtfishman:
My main suggestion would be choosing an ordering for the vertices, since that appears to be arbitrary with the current implementation.

What about sorting from largest to smallest eccentricity (https://juliagraphs.org/Graphs.jl/dev/algorithms/distance/), with the intuition that you move inward from the periphery of the graph? Unfortunately it seems like `eccentricity` is broken for NamedGraphs but that should be easy to fix.

src/traversals/dfs.jl
```diff
@@ -25,3 +25,32 @@ end
...
+#Given a graph, split it into its connected components, construct a spanning tree over each of them
+# and take the union (adding in edges between the trees to recover a connected graph)
+function spanning_forest(g::AbstractNamedGraph)
+  components = connected_components(g)
+  return reduce(
+    union,
+    NamedGraph[
+      undirected_graph(bfs_tree(g[g_comp], first(g_comp))) for g_comp in components
+    ],
+  )
```
mtfishman, line 38: Can use a comprehension instead of allocating a vector, i.e.
```julia
reduce(union, (undirected_graph(bfs_tree(g[g_comp], first(g_comp))) for g_comp in components))
```
Also, I found it confusing calling it `g_comp` since I thought they were graphs but I realized they are lists of vertices, probably good to indicate that with a name like `component_vertices`.

mtfishman:
Here's an example of sorting vertices from largest to smallest eccentricity:
```julia
julia> g = named_grid((4, 4));

julia> sort(Dictionary(vertices(g), eccentricity(g.parent_graph)); rev=true)
16-element Dictionary{Tuple{Int64, Int64}, Int64}
 (1, 1) │ 6
 (4, 1) │ 6
 (1, 4) │ 6
 (4, 4) │ 6
 (2, 1) │ 5
 (3, 1) │ 5
 (1, 2) │ 5
 (4, 2) │ 5
 (1, 3) │ 5
 (4, 3) │ 5
 (2, 4) │ 5
 (3, 4) │ 5
 (2, 2) │ 4
 (3, 2) │ 4
 (2, 3) │ 4
 (3, 3) │ 4
```
which seems like a reasonable ordering. For grids that looks like it could give a minimal number of spanning trees.

JoeyT1994:
Yeah there is no randomness in the tests? It was just picking the first vertex in the list of the graph's vertices to do a spanning tree over. Switching to eccentricity sounds reasonable. On my end `eccentricities(g::NamedGraph)` seems to work. The failing test seems to come from `test_namedgraph.jl` as opposed to the new ones I introduced?

JoeyT1994:
This doesn't cover edges in both directions and the arboricity is not really defined for Directed graphs. Hence I have added an assertion that the graph is undirected. I think if the user really wanted to cover a directed graph they could split it into two undirected graphs and pass them to the function and concatenate the result.

JoeyT1994:
Okay this is breaking because of `Graphs.jl` changing from 1.8.0 to 1.9.0 (https://github.com/JuliaGraphs/Graphs.jl/releases/tag/v1.9.0).  Those changes have broken the relevant tests (as opposed to anything in this PR), you can see that they directly effect functions like `topological_sort_by_dfs` in the change list.

JoeyT1994:
Looks like the failing tests are just because they are testing for a specific solution and the modified traversal algorithms find different, also valid, solutions. I am fixing that.

JoeyT1994:
Okay these are all passing now as I made the failing tests just test they get a correct solution as opposed to a specific solution with specific orderings etc... which can change depending on algorithmic implementations

JoeyT1994:
Okay I updated the interface a bit like we discussed. Although I didn't do it in terms of  the`algorithm` type because is that not an `ITensor` feature which we don't want to import here? Unless I am missing something.

src/traversals/dfs.jl
```diff
@@ -25,3 +25,36 @@ end
...
+  while !isempty(remaining_edges)
+    g_reduced = rem_edges(g, edges_collected)
+    g_reduced_spanning_forest = spanning_forest(g_reduced; tree_constructor)
+    push!(edges_collected, edges(g_reduced_spanning_forest)...)
+    push!(forests, g_reduced_spanning_forest)
+    setdiff!(remaining_edges, edges(g_reduced_spanning_forest))
+  end
+
+  return forests
+end
```
mtfishman, line 60: Guess this should be removed?

mtfishman:
> Okay I updated the interface a bit like we discussed. Although I didn't do it in terms of the`algorithm` type because is that not an `ITensor` feature which we don't want to import here? Unless I am missing something.

Good point.

Instead we could just define types, for example:
```julia
abstract type SpanningTreeAlgorithm end

struct BFS <: SpanningTreeAlgorithm end
struct RandomBFS <: SpanningTreeAlgorithm end
struct DFS <: SpanningTreeAlgorithm end

default_spanning_tree_alg() = BFS()
spanning_tree(g::AbstractNamedGraph, src) = spanning_tree(g, src, default_spanning_tree_alg())
spanning_tree(g::AbstractNamedGraph, alg::SpanningTreeAlgorithm=default_spanning_tree_alg()) = spanning_tree(g, default_root_vertex(), alg)

# Implementations
spanning_tree(g::AbstractNamedGraph, src, ::BFS) = undirected_graph(bfs_tree(g, src))
```

src/traversals/trees_and_forests.jl
```diff
@@ -0,0 +1,43 @@
...
+    return undirected_graph(dfs_tree(g, root_vertex))
+  else
+    error("Algorithm not current supported")
+  end
+end
+
+#Given a graph, split it into its connected components, construct a spanning tree over each of them
+# and take the union.
+function spanning_forest(
+  g::AbstractNamedGraph; spanning_tree_function=g -> spanning_tree(g)
```
mtfishman, line 21:
```suggestion
  g::AbstractNamedGraph; spanning_tree=spanning_tree
```
i.e. shorten the kwarg from `spanning_tree_function` to `spanning_tree`, and also you can just use `f` instead of `x -> f(x)` in general.

src/traversals/trees_and_forests.jl
```diff
@@ -0,0 +1,43 @@
...
+)
+  return reduce(union, (spanning_tree_function(g[vs]) for vs in connected_components(g)))
+end
+
+#Given an undirected graph g with vertex set V, build a set of forests (each with vertex set V) which covers all edges in g
+# (see https://en.wikipedia.org/wiki/Arboricity) We do not find the minimum but our tests show this algorithm performs well
+function build_forest_cover(
```
mtfishman, line 28:
```suggestion
function forest_cover(
```

src/traversals/trees_and_forests.jl
```diff
@@ -0,0 +1,43 @@
...
+  g::AbstractNamedGraph; spanning_tree_function=g -> spanning_tree(g)
```
mtfishman, line 29:
```suggestion
  g::AbstractNamedGraph; spanning_tree=spanning_tree
```

mtfishman:
See https://github.com/JuliaGraphs/Graphs.jl/blob/v1.9.0/src/Experimental/Traversals/bfs.jl for some inspiration for the proposed interface.

JoeyT1994:
Okay I have done it by defining types as you suggested. Let me know what you think about that? I put the algorithm as the first argument and left the root_vertex as keyword.

Also seeming as I am asserting the graph is not directed when calling `spanning_tree` should I make the trait functions instead?

mtfishman:
Yeah makes sense to make them trait functions.

JoeyT1994:
Okay I changed them to traits

test/test_trees_and_forests.jl
```diff
@@ -0,0 +1,46 @@
+using Test
+using Graphs
+using NamedGraphs
+using NamedGraphs:
+  hexagonal_lattice_graph, triangular_lattice_graph, forest_cover, spanning_tree
+
+@testset "Test Spanning Trees" begin
```
mtfishman, line 7: Something neat you can do is merge the for loops and the testset:
```julia
@testset "Test Spanning Trees $g, $alg" for g in [...], alg in [...]
  # Test code
end
```

src/traversals/trees_and_forests.jl
```diff
@@ -0,0 +1,54 @@
...
+
+@traitfn function spanning_tree(
+  ::DFS, g::AbstractNamedGraph::(!IsDirected); root_vertex=default_root_vertex(g)
+)
+  return undirected_graph(dfs_tree(g, root_vertex))
+end
+
+#Given a graph, split it into its connected components, construct a spanning tree over each of them
+# and take the union.
+function spanning_forest(g::AbstractNamedGraph; spanning_tree_function=spanning_tree)
```
mtfishman, line 35: I would lean toward just calling it `spanning_tree` instead of `spanning_tree_function`, we could document that it accepts a function.

test/test_trees_and_forests.jl
```diff
@@ -4,43 +4,30 @@ using NamedGraphs
...
-  end
+gs = [
+  ("Chain", named_grid((6, 1))),
+  ("Cubic Lattice", named_grid((3, 3, 3))),
+  ("Hexagonal Grid", hexagonal_lattice_graph(6, 6)),
+  ("Comb Tree", named_comb_tree((4, 4))),
+  ("Square lattice", named_grid((10, 10))),
+  ("Triangular Grid", triangular_lattice_graph(5, 5; periodic=true)),
+];
+algs = [NamedGraphs.BFS(), NamedGraphs.DFS(), NamedGraphs.RandomBFS()];
```
mtfishman, line 15: If someone `include`s this test file this introduces global variables which can cause subtle bugs (like if another code makes use of a variable `algs`, this could overwrite that variable).

We could wrap these in a module like:
```julia
module TestTreesAndForests
gs = [
  ("Chain", named_grid((6, 1))),
  ("Cubic Lattice", named_grid((3, 3, 3))),
  ("Hexagonal Grid", hexagonal_lattice_graph(6, 6)),
  ("Comb Tree", named_comb_tree((4, 4))),
  ("Square lattice", named_grid((10, 10))),
  ("Triangular Grid", triangular_lattice_graph(5, 5; periodic=true)),
]
algs = [NamedGraphs.BFS(), NamedGraphs.DFS(), NamedGraphs.RandomBFS()]
end
```
then use `TestTreesAndForests.gs`, `TestTreesAndForests.algs` in the code below.

mtfishman: That kind of thing has caused me some serious headaches with bugs that are really hard to track down.

JoeyT1994: Good point. Yes I also have had a nightmare with bugs like that. Fixed it now I think

src/traversals/trees_and_forests.jl
```diff
@@ -0,0 +1,54 @@
...
+struct RandomBFS <: SpanningTreeAlgorithm end
+struct DFS <: SpanningTreeAlgorithm end
+
+default_spanning_tree_alg() = BFS()
+
+default_root_vertex(g) = last(findmax(eccentricities(g)))
+
+function spanning_tree(g::AbstractNamedGraph; root_vertex=default_root_vertex(g))
+  return spanning_tree(default_spanning_tree_alg(), g; root_vertex)
+end
```
mtfishman, line 13:
```suggestion
function spanning_tree(g::AbstractNamedGraph; alg=default_spanning_tree_alg(), root_vertex=default_root_vertex(g))
  return spanning_tree(alg, g; root_vertex)
end
```
Then users can pass the `alg` as a keyword argument, a bit nicer than having to pass it as the first argument in terms of an external interface.

test/test_trees_and_forests.jl
```diff
@@ -0,0 +1,38 @@
...
+  ("Square lattice", named_grid((10, 10))),
+  ("Triangular Grid", triangular_lattice_graph(5, 5; periodic=true)),
+]
+algs = [NamedGraphs.BFS(), NamedGraphs.DFS(), NamedGraphs.RandomBFS()]
+end
+
+@testset "Test Spanning Trees $g_string, $alg" for (g_string, g) in TestTreesAndForests.gs,
+  alg in TestTreesAndForests.algs
+
+  s_tree = spanning_tree(alg, g)
```
mtfishman, line 23:
```suggestion
  s_tree = spanning_tree(g; alg)
```
Let's consider the kwarg syntax to be the external API, so the tests should test that interface.

mtfishman:
Looks like it just needs a formatting fix and then good to merge. Thanks for sticking with the comments!

Issue #41, opened 2023-12-06
Introduce a PartitionedGraph Type


JoeyT1994:
We should introduce a `PartitionedGraph` structure. Ideally this should be separate from `NamedGraphs` and just be an abstract graph type with the relevant functionality. For now though we will make a `PartitionedGraphs.jl` submodule within `NamedGraphs.jl`.

The general structure could look like

```
abstract type AbstractPartitionedGraph{V,PV} <: AbstractGraph{V} end

struct PartitionedGraph{V,PV,G<:AbstractGraph{V},PG<:AbstractGraph{PV}} <: AbstractPartitionedGraph{PV}
  graph::G
  partioned_graph::PG
  partition_vertices::Dictionary{PV,Vector{V}}
  partition::Dictionary{V,PV}
end
```

such that a partitioned graph essentially consists of two graphs: the original graph `G` and the partitioned graph `PG`. The structure then also contains a map (and its reverse) between the subsets of vertices of `G` which consistute the vertices of `PG`. 

We then need to define functionality for this structure. It would make sense to define standard `Graphs.jl` calls on a `PartitionedGraph` to pass to the original graph `G`. For instance `vertices(pg::PartitionedGraph) = vertices(pg.G)` Then we define partitioned versions of those calls which operate on `PG` instead. E.g. `partitioned_vertices(pg::PartitionedGraph) = vertices(pg.PG)`. 
Other examples of such functions might be `has_vertex` versus `partitioned_has_vertex`, `is_tree` versus `partitioned_is_tree` etc. 

Helpful functions might then be determining whether an edge of `G` is within a partition or between partitions. And which partitions.
E.g
```
function edge_partitions(e::AbstractEdge, PG::PartitionedGraph)
   return find_key(PG.partition, src(e)), find_key(PG.partition, dst(e))
end

function edge_in_partition(e::AbstractEdge, PG::PartitionedGraph)
  return all_equal(edge_partitions(e, PG))
end
```

Another question is whether we define types such as `PartitionedVertex` and `PartitionedEdge` so that calls such as `has_vertex(PG, PartitionedVertex(1))` pass directly to `partitioned_has_vertex`. This is mostly for convenience.
    
@mtfishman

JoeyT1994:
We should also define default constructors for a `PartitionedGraph`.

[...]

PR #44, opened 2023-12-13
Partitioned graphs

JoeyT1994:
This PR adds support for partitioned graphs (see https://en.wikipedia.org/wiki/Graph_partition and the relevant issue #41)

Specifically an `AbstractPartitionedGraph pg` is a structure which contains: 

- The graph which was partitioned `g` (can be accessed via `graph(pg)`)
- The new, simpler graph, `g_partitioned` (can be accessed via `partitioned_graph(pg)`)
- A map from each of the vertices `pv` of `g_partitioned` to the corresponding set of vertices in `g`. This information can be accessed via `vertices(pg, pv)`.
- A map from each vertex `v` in `g` to its corresponding vertex in `g_partitioned`. This can be accessed with `partition_vertex(pg, v)`

As the `AbstractPartitionedGraph` is a subtype of the `AbstractNamedGraph`, `NamedGraphs` functions immediately can be applied to it. Most functions will simply forward to the underlying graph `g`, i.e. `is_tree(pg) = is_tree(graph(g))`.
However, for functions which actually modify the graph or return a `NamedGraph` type we modify the whole `partitioned_graph`.
E.g. `rem_vertex!(pg, v)` will remove `vertex` from `graph(pg)` and also modify `partitioned_graph(pg)` and the corresponding vertex maps appropriately.

We also introduce the `AbstractPartitionVertex` and `AbstractPartitionEdge` type to overload a function like `rem_vertex` so that `rem_vertex!(pg, pv::AbstractPartitionVertex)` exists and will remove `pv` from `partitioned_graph(pg)` and then update `graph(pg)` and the relevant maps correspondingly.

This PR also introduces the concrete `PartitionedGraph` type which is structured explicitly as:
```
struct PartitionedGraph{V,PV,G<:AbstractNamedGraph{V},PG<:AbstractNamedGraph{PV}} <:
       AbstractPartitionedGraph{V,PV}
  graph::G
  partitioned_graph::PG
  partition_vertices::Dictionary{PV,Vector{V}}
  partition_vertex::Dictionary{V,PV}
end
```
and operates as expected.
Some example code that can now be written with this PR is:

```
  using NamedGraphs
  nx, ny = 5, 5
  g = named_grid((nx, ny))

  partitions = [[(i, j) for j in 1:ny] for i in 1:nx]
  pg = PartitionedGraph(g, partitions)
  @assert is_path_graph(pg.partitioned_graph)
  ```
  which partitions the grid by its rows resulting in `pg.pg.partitioned_graph' being a 5-site path graph.
  
  Extensive testing is included in `tests/test_partitionedgraph.jl` and an example is included in `examples/partitioning.jl`. The partitioning libraries `KaHyPar.jl` and `Metis.jl` can be called to build a `PartitionedGraph` from a `NamedGraph` `g` based on specifying some parameter like the number of partitions and minimising some cost function like the number of edges between partitions. This is the default cost function.
  
 E.g. `PartitionedGraph(g; npartitions, backend = "KaHyPar")` will construct a `PartitionedGraph` with `npartitions` and attempt to partition in such a way as to minimise the number of edges between partitions.
 
 Lots of the code here supercedes the code in `src/partition.jl` of `ITensorNetworks.jl` and thus that will be removed in a near term PR.

### CODE REVIEW

File: src/abstractnamedgraph.jl [hunk 1]
```diff
@@ -381,10 +381,7 @@ function union(graph1::AbstractNamedGraph, graph2::AbstractNamedGraph)
   return union_graph
 end
 
-function add_vertex!(graph::AbstractNamedGraph, vertex)
-  if vertex ∈ vertices(graph)
-    throw(ArgumentError("Duplicate vertices are not allowed"))
-  end
+function insert_vertex!(graph::AbstractNamedGraph, vertex)
```
mtfishman, line 384: I was picturing this function would throw an error if `vertex ∈ vertices(graph)`.

mtfishman: Like how `insert!` works in `Dictionaries.jl`.

JoeyT1994: So more like this?

```
function set_vertex!(graph::AbstractNamedGraph, vertex)
  add_vertex!(parent_graph(graph))
  # Update the vertex list
  push!(vertices(graph), vertex)
  # Update the reverse map
  # TODO: Make this more generic
  insert!(graph.vertex_to_parent_vertex, vertex, last(parent_vertices(graph)))
  return graph
end

function insert_vertex!(graph::AbstractNamedGraph, vertex)

  if vertex ∈ vertices(graph)
    error("Duplicate vertices not allowed!")
  else
    set_vertex!(graph, vertex)
  end

end

function add_vertex!(graph::AbstractNamedGraph, vertex)
  if vertex ∈ vertices(graph)
    return false
  else
    set_vertex!(graph, vertex)
    return true
  end
end
```

JoeyT1994: And similarly for `unset_vertex!`, `delete_vertex!` and `rem_vertex!`?

mtfishman: Yeah looks right. Makes sense to add those as well.

I guess we could define `set_vertex!`/`unset_vertex!` and `insert_vertex!`/`delete_vertex!` for general `AbstractGraph`, though could be defined using `add_vertex!`/`rem_vertex!`.

JoeyT1994: Do you mean that we could define `insert_vertex!` and `add_vertex!` for an `AbstractGraph` type? These would then rely on an implementation of `set_vertex!` for the concrete type? 

As `set_vertex!` seems to be the one which is really specific to the implementation of the graph?

mtfishman: Yeah I think there is some question about what the minimal interface would be, i.e. by default which ones are expected to be defined in terms of which other ones. I just proposed `add_vertex!` as the required core function since that is already expected to be defined of `AbstractGraph` according to the `Graphs.jl` interface.

src/Graphs/partitionedgraphs/partitionedgraph.jl
```diff
@@ -0,0 +1,167 @@
...
+end
+
+function PartitionedGraph(
+  g::AbstractNamedGraph{V};
+  npartitions=nothing,
+  nvertices_per_partition=nothing,
+  backend=current_partitioning_backend(),
+  kwargs...,
+) where {V}
+  partition_vertices = partition(
```
mtfishman, line 71: I wonder if this should be called `partition_vertices(...)` instead of just `partition(...)` to clarify it outputs vertices.

`partition(...)` could be a shorthand notation for creating a `PartitionedGraph`.

src/Graphs/partitionedgraphs/partitionedgraph.jl
```diff
@@ -0,0 +1,167 @@
...
+    g; npartitions, nvertices_per_partition, backend, kwargs...
+  )
+  return PartitionedGraph(g, partition_vertices)
+end
+
+#Needed for interface
+partitioned_graph(pg::PartitionedGraph) = getfield(pg, :partitioned_graph)
+graph(pg::PartitionedGraph) = getfield(pg, :graph)
+partition_vertices(pg::PartitionedGraph) = getfield(pg, :partition_vertices)
+partition_vertex(pg::PartitionedGraph) = getfield(pg, :partition_vertex)
```
mtfishman, line 81: These are tricky to name.

Maybe `partitioned_vertices` would be better than `partition_vertices`, since it is a collection of the partitioned vertices (i.e. stores how the vertices of the underlying graph are partitioned).

The best I can think of as an alternative to `partition_vertex` is `which_partition` or maybe `vertex_to_partition_vertex`.

src/Graphs/partitionedgraphs/partitionedgraph.jl
```diff
@@ -0,0 +1,167 @@
...
+function copy(pg::PartitionedGraph)
+  return PartitionedGraph(
+    copy(graph(pg)),
+    copy(partitioned_graph(pg)),
+    copy_keys_values(partition_vertices(pg)),
+    copy_keys_values(partition_vertex(pg)),
+  )
+end
+
+function add_to_vertex_map!(pg::PartitionedGraph, vertex, partition_vertex::PartitionVertex)
```
mtfishman, line 123: I would go with `insert_vertex_map!` or `set_vertex_map!` (depending on if it errors if the vertex exists already).

src/Graphs/partitionedgraphs/partitionedgraph.jl
```diff
@@ -0,0 +1,167 @@
...
+  if pv ∉ keys(partition_vertices(pg))
+    insert!(partition_vertices(pg), pv, [vertex])
+  else
+    pg.partition_vertices[pv] = unique(vcat(vertices(pg, partition_vertex), [vertex]))
+  end
+
+  return insert!(NamedGraphs.partition_vertex(pg), vertex, pv)
+end
+
+function rem_from_vertex_map!(pg::PartitionedGraph, vertex)
```
mtfishman, line 134: I would go with `delete_vertex_map!` or `unset_vertex_map!` (depending on if it errors if the vertex doesn't exist already).

src/Graphs/partitionedgraphs/partitionedgraph.jl
JoeyT1994: Yeah `partitioned_vertices` sounds a bit better. Have also gone with `which_partition` for now.

src/Graphs/partitionedgraphs/abstractpartitionedgraph.jl
```diff
@@ -0,0 +1,168 @@
+abstract type AbstractPartitionedGraph{V,PV} <: AbstractNamedGraph{V} end
+
+#Needed for interface
+partitioned_graph(pg::AbstractPartitionedGraph) = not_implemented()
+graph(pg::AbstractPartitionedGraph) = not_implemented()
```
mtfishman, line 5: Maybe `unpartitioned_graph`?

src/Graphs/partitionedgraphs/abstractpartitionedgraph.jl
```diff
@@ -0,0 +1,168 @@
...
+    return false
+  end
+
+  for field in fieldnames(typeof(pg1))
+    if getfield(pg1, field) != getfield(pg2, field)
+      return false
+    end
+  end
+  return true
+end
```
mtfishman, line 168: This doesn't seem quite right, since it is ok for the field names to be different as long as the graph and vertex partitioning is equal.

So it seems best to compare `unpartitioned_graph(pg1) == unpartitioned_graph(pg2) && partitioned_vertices(pg1) == partitioned_vertices(pg2)`, or something like that. Comparing the partitioned vertices may be a bit more subtle, perhaps `issetequal(partition_vertices(pg1), partition_vertices(pg2)) && all(pv -> issetequal(vertices(pg1, [pv]), vertices(pg2, [pv]))), partition_vertices(pg1))`.

src/Graphs/partitionedgraphs/partitionedgraph.jl
```diff
@@ -0,0 +1,167 @@
...
+  return PartitionedGraph(g, pg, partitioned_vertices, which_partition)
+end
+
+function PartitionedGraph(
+  g::AbstractNamedGraph{V}, partitioned_vertices::Dictionary{PV,Vector{V}}
+) where {V,PV}
+  return PartitionedGraph{V,PV,NamedGraph{V},NamedGraph{PV}}(g, partitioned_vertices)
+end
+
+function PartitionedGraph{V,Int64,G,NamedGraph{Int64}}(
```
mtfishman, line 39: I don't see why this needs to be specialized to `Int64`.

test/test_partitionedgraph.jl
```diff
@@ -0,0 +1,111 @@
+using Test
+using NamedGraphs
```
mtfishman, line 2: Is this needed?

src/Graphs/partitionedgraphs/abstractpartitionedgraph.jl
```diff
@@ -0,0 +1,168 @@
...
+  return rem_edges!(pg, edges(pg, parent(partition_edge)))
+end
+
+function rem_edge(pg::AbstractPartitionedGraph, partition_edge::AbstractPartitionEdge)
+  pg_new = copy(pg)
+  rem_partition_edge!(pg_new, partition_edge)
+  return pg_new
+end
+
+function rem_partition_edges!(
```
mtfishman, line 73: Can't this just be `rem_edges!`?

src/Graphs/partitionedgraphs/abstractpartitionedgraph.jl
```diff
@@ -0,0 +1,168 @@
...
+  pg::AbstractPartitionedGraph, partition_edges::Vector{AbstractPartitionEdge}
```
mtfishman, line 74:
```suggestion
  pg::AbstractPartitionedGraph, partition_edges::Vector{<:AbstractPartitionEdge}
```

src/Graphs/partitionedgraphs/abstractpartitionedgraph.jl
```diff
@@ -0,0 +1,168 @@
...
+  add_vertex!(graph(pg), vertex)
+  add_vertex!(partitioned_graph(pg), parent(partition_vertex))
+
+  return insert_to_vertex_map!(pg, vertex, partition_vertex)
+end
+
+function add_vertices!(
+  pg::AbstractPartitionedGraph,
+  vertices::Vector,
+  partition_vertices::Vector{V} where {V<:AbstractPartitionVertex},
```
mtfishman, line 102:
```suggestion
  partition_vertices::Vector{<:AbstractPartitionVertex},
```

src/Graphs/partitionedgraphs/abstractpartitionedgraph.jl
```diff
@@ -0,0 +1,168 @@
...
+end
+
+function rem_vertex(pg::AbstractPartitionedGraph, partition_vertex::AbstractPartitionVertex)
+  pg_new = copy(pg)
+  rem_vertex!(pg_new, partition_vertex)
+  return pg_new
+end
+
+function rem_vertices!(
+  pg::AbstractPartitionedGraph, partition_vertices::Vector{AbstractPartitionVertex}
```
mtfishman, line 138:
```suggestion
  pg::AbstractPartitionedGraph, partition_vertices::Vector{<:AbstractPartitionVertex}
```

src/Graphs/partitionedgraphs/abstractpartitionedgraph.jl
```diff
@@ -0,0 +1,168 @@
...
+)
+  for pv in partition_vertices
+    rem_vertex!(pg, pv)
+  end
+end
+
+function rem_vertices(
+  pg::AbstractPartitionedGraph, partition_vertices::Vector{AbstractPartitionVertex}
```
mtfishman, line 146:
```suggestion
  pg::AbstractPartitionedGraph, partition_vertices::Vector{<:AbstractPartitionVertex}
```

src/Graphs/partitionedgraphs/partitionedgraph.jl
```diff
@@ -0,0 +1,167 @@
+struct PartitionedGraph{V,PV,G<:AbstractNamedGraph{V},PG<:AbstractNamedGraph{PV}} <:
```
mtfishman, line 1: I don't think these need to be constrained to `AbstractNamedGraph`, for example I could imagine they might be `DataGraph`.

src/Graphs/partitionedgraphs/partitionedgraph.jl
```diff
@@ -0,0 +1,167 @@
...
+function PartitionedGraph{V,V,G,G}(vertices::Vector{V}) where {V,G<:NamedGraph{V}}
+  return PartitionedGraph(NamedGraph{V}(vertices), [[v] for v in vertices])
+end
+
+function PartitionedGraph(vertices::Vector{V}) where {V}
+  return PartitionedGraph{V,V,NamedGraph{V},NamedGraph{V}}(vertices)
+end
+
+function PartitionedGraph(
+  g::AbstractNamedGraph{V};
```
mtfishman, line 65:
```suggestion
  g::AbstractNamedGraph;
```
`V` doesn't seem to be used.

src/Graphs/partitionedgraphs/partitionedgraph.jl
JoeyT1994: Sure, I will change to `AbstractGraph`!

src/Graphs/partitionedgraphs/abstractpartitionedgraph.jl
JoeyT1994: Hmmm yeah it does seem difficult to compare.

What about just asserting `unpartitioned_graph(pg1) == unpartitioned_graph(pg2)', asserting `vertices(pg1) == vertices(pg2)` and then asserting that each vertex of the unpartitionedgraphs lies in the same partition, I.e. `which_partition(pg1, v) == which_partition(pg2, v)` for all `v`.

JoeyT1994: Obviously to account for asserting things like that two `partitionedITensorNetworks` are equal we would have to overload the function to  check the data as well.

mtfishman: That seems like a good approach. Probably checking `vertices(pg1) == vertices(pg2)` (though probably would need to check with set equality since they might be out of order from each other) isn't necessary if we check `unpartitioned_graph(pg1) == unpartitioned_graph(pg2)`. Also, if we additionally check `partitioned_graph(ph1) == partitioned_graph(pg2)` that should cover comparing the data as well, assuming data is checked in equality of data graphs.

src/Graphs/partitionedgraphs/abstractpartitionedgraph.jl
```diff
@@ -0,0 +1,173 @@
...
+function rem_edges!(
+  pg::AbstractPartitionedGraph, partition_edges::Vector{<:AbstractPartitionEdge}
+)
+  for pe in partition_edges
+    rem_edge!(pg, pe)
+  end
+end
+
+function rem_edges(
+  pg::AbstractPartitionedGraph, partition_edges::Vector{AbstractPartitionEdge}
```
mtfishman, line 85:
```suggestion
  pg::AbstractPartitionedGraph, partition_edges::Vector{<:AbstractPartitionEdge}
```

src/Graphs/partitionedgraphs/abstractpartitionedgraph.jl
JoeyT1994: Makes sense! I have now made it check both the `partitioned_graph` and `unpartitioned_graph`, along with the vertex groupings.

src/requires/kahypar.jl
```diff
@@ -0,0 +1,33 @@
+set_partitioning_backend!(Backend"KaHyPar"())
+
+# https://github.com/kahypar/KaHyPar.jl/issues/20
+KaHyPar.HyperGraph(g::SimpleGraph) = incidence_matrix(g)
+
+"""
+    partition(::Backend"KaHyPar", g::Graph, npartiations::Integer; objective="edge_cut", alg="kway", kwargs...)
```
mtfishman, line 7:
```suggestion
    partition_vertices(::Backend"KaHyPar", g::Graph, npartiations::Integer; objective="edge_cut", alg="kway", kwargs...)
```

src/requires/metis.jl
```diff
@@ -0,0 +1,33 @@
+set_partitioning_backend!(Backend"Metis"())
+
+"""
+    partition(::Backend"Metis", g::AbstractGraph, npartitions::Integer; alg="recursive")
```
mtfishman, line 4:
```suggestion
    partition_vertices(::Backend"Metis", g::AbstractGraph, npartitions::Integer; alg="recursive")
```

test/test_partitionedgraph.jl
```diff
@@ -0,0 +1,110 @@
...
+using Graphs
+
+@testset "Test Partitioned Graph Constructors" begin
+  nx, ny = 10, 10
+  g = named_grid((nx, ny))
+
+  partitions = [[(i, j) for j in 1:ny] for i in 1:nx]
+  pg = PartitionedGraph(g, partitions)
+  @test vertextype(pg.partitioned_graph) == Int64
+  @test vertextype(pg.graph) == vertextype(g)
```
mtfishman, line 23: Better to use function accessors in tests rather than accessing fields with `.`, since accessing fields should be considered an internal detail.

mtfishman: Same for the rest of the tests.

test/test_partitionedgraph.jl
```diff
@@ -0,0 +1,110 @@
...
+  pv = PartitionVertex(5)
+  v_set = vertices(pg, pv)
+  edges_involving_v_set = filter(
+    e -> !isempty(intersect(v_set, [src(e), dst(e)])), edges(pg)
+  )
+
+  #Strip the middle column from pg via the partitioned graph vertex, and make a new pg
+  pg_stripped = NamedGraphs.rem_vertex(pg, pv)
+  @test !is_connected(pg_stripped.graph) && !is_connected(pg_stripped.partitioned_graph)
+  @test !haskey(pg_stripped.partitioned_vertices, parent(pv))
```
mtfishman, line 50: I think this isn't a good idea for a test since it is reaching into internal details of the type which in principle could change. The test below this one is the proper external interface so is sufficient for testing this functionality.

test/test_partitionedgraph.jl
```diff
@@ -0,0 +1,110 @@
...
+  g = named_grid((nx, ny))
+
+  partitions = [[(i, j) for j in 1:ny] for i in 1:nx]
+  pg = PartitionedGraph(g, partitions)
+
+  pv = PartitionVertex(5)
+  v_set = vertices(pg, pv)
+  edges_involving_v_set = filter(
+    e -> !isempty(intersect(v_set, [src(e), dst(e)])), edges(pg)
+  )
```
mtfishman, line 45: Could this use `boundary_edges` instead?

test/test_partitionedgraph.jl
```diff
@@ -0,0 +1,110 @@
...
+  @test !haskey(pg.partitioned_vertices, parent(pv))
+  @test !has_vertex(pg, pv)
+
+  #Test both are the same
+  @test pg == pg_stripped
+
+  #Add the column back to the in place graph
+  NamedGraphs.add_vertices!(pg, v_set, pv)
+  NamedGraphs.add_edges!(pg, edges_involving_v_set)
+  @test is_connected(pg.graph) && is_path_graph(pg.partitioned_graph)
```
mtfishman, line 65: Probably good to test some more basic properties as well, like number of vertices and edges in the graph and partitioned graph, checking that the partitioned graph is a tree, etc.

mtfishman: Same thing for other places in the tests.

test/test_partitionedgraph.jl
```diff
@@ -0,0 +1,110 @@
...
+  @test haskey(pg.partitioned_vertices, parent(pv))
+  @test has_vertex(pg, pv)
+end
+
+@testset "Test Partitioned Graph Subgraph Functionality" begin
+  n, z = 20, 4
+  nvertices_per_partition = 4
+  g = NamedGraph(random_regular_graph(n, z))
+  npartitions = _npartitions(g, nothing, nvertices_per_partition)
```
mtfishman, line 74: Why use `_npartitions` here? That seems like an internal function.  Basically I don't really understand what this code is doing so it would be better to simplify it and try to avoid using internal functions.

test/test_partitionedgraph.jl
```diff
@@ -0,0 +1,110 @@
...
+  partitions = [
+    [nvertices_per_partition * (i - 1) + j for j in 1:nvertices_per_partition] for
+    i in 1:npartitions
+  ]
+  pg = PartitionedGraph(g, partitions)
+
+  pg_1 = subgraph(pg, partitions[1])
+  pg_2 = subgraph(pg, [PartitionVertex(1)])
```
mtfishman, line 82: I think it could be ok to allow the syntax `subgraph(pg, PartitionVertex(1))` as well. What I could imagine is:
```julia
# Outputs a `PartitionedGraph`, happens to have one partition but generalizes to more partitions
subgraph(pg, [PartitionVertex(1)])
# Outputs a `NamedGraph`
subgraph(pg, PartitionVertex(1))
```

test/test_partitionedgraph.jl
JoeyT1994: The code is partitioning up a random regular graph into some number of partitions and then testing the functionality of `subgraph(pg::PartitionedGraph, )` on it. I will simplify to improve readability

mtfishman: I see, now I understand. I was wondering why you were calling `_npartitions` and couldn't just call `partition_vertices`, but I realize now you are just calling `_npartitions` to get the number of partitions and then you are using that to perform a custom partitioning of the vertices.

At the very least it seems like you can just directly use:
```julia
npartitions = nv(g) ÷ nvertices_per_partition
```
instead of using the internal function `_npartitions`, the rest seems ok.

To actually do the vertex partitioning, since I guess for a random graph it is basically arbitrary anyway, if I understand you just need to split the vector of vertices into parts or chunks and there should be various ways of doing that with Base Julia or some Julia package, see for example https://discourse.julialang.org/t/split-vector-into-n-potentially-unequal-length-subvectors/73548.

test/test_partitionedgraph.jl
JoeyT1994: That makes sense! I will add it

test/test_partitionedgraph.jl
JoeyT1994: Good point. Changed it.

test/test_partitionedgraph.jl
JoeyT1994: Sure, I have added those kind of tests throughout.

examples/partitioning.jl
```diff
@@ -0,0 +1,16 @@
...
+
+g = named_grid((4, 4))
+npartitions = 4
+
+pg_kahypar = PartitionedGraph(g; npartitions, backend="KaHyPar")
+pg_metis = PartitionedGraph(g; npartitions, backend="Metis")
+
+@show length(vertices(pg_kahypar.partitioned_graph)) ==
+  length(vertices(pg_metis.partitioned_graph)) ==
+  npartitions
```
mtfishman, line 13: Should use `partitioned_graph(...)` function interface instead.

examples/partitioning.jl
```diff
@@ -0,0 +1,16 @@
...
+
+@show typeof(pg_kahypar) <: PartitionedGraph
+@show typeof(pg_metis) <: PartitionedGraph
```
mtfishman, line 16: I prefer the style `pg_kahypar isa PartitionedGraph`.

src/Graphs/partitionedgraphs/partitionedgraph.jl
```diff
@@ -0,0 +1,172 @@
...
+  npartitions=nothing,
+  nvertices_per_partition=nothing,
+  backend=current_partitioning_backend(),
+  kwargs...,
+)
+  partitioned_vertices = partition_vertices(
+    g; npartitions, nvertices_per_partition, backend, kwargs...
+  )
+  return PartitionedGraph(g, partitioned_vertices)
+end
```
mtfishman, line 73: Designing constructors can be tricky. It seems like these can be simplified by making them a bit more generic.

Here's a proposal:
```julia
# Can work when `partitioned_vertices` is a `Vector`, `Dictionary`, etc. (some collection where the keys are interpreted as the partitioned vertices and the values are the vertices in that partition).
# Also assumes `g` is NamedGraph-like, but we can't constrain that since it could be a `DataGraph`.
function PartitionedGraph(
  g::AbstractGraph, partitioned_vertices
)
  pvs = keys(partitioned_vertices)
  pg = NamedGraph(pvs)
  # Rest of code
  return PartitionedGraph(g, pg, partitioned_vertices, which_partition)
end

# Can work with `partitioned_vertices` as a `Vector` or `Dictionary`
function PartitionedGraph(partitioned_vertices)
  return PartitionedGraph(NamedGraph(keys(partitioned_vertices)), partitioned_vertices)
end

function PartitionedGraph(
  g::AbstractGraph;
  npartitions=nothing,
  nvertices_per_partition=nothing,
  backend=current_partitioning_backend(),
  kwargs...,
)
  partitioned_vertices = partition_vertices(
    g; npartitions, nvertices_per_partition, backend, kwargs...
  )
  return PartitionedGraph(g, partitioned_vertices)
end
```
Not sure if that is sufficient or correct (not tested) but hopefully you get the idea, the goal is to make fewer constructors that are more generic and avoid over-specifying type parameters and constraining input types as much as possible.

JoeyT1994: Okay that makes sense. I think if I remove the type constraint on the constructor so that it's just

```
struct PartitionedGraph{V,PV,G<:AbstractGraph{V},PG<:AbstractGraph{PV}} <:
       AbstractPartitionedGraph{V,PV}
  graph::G
  partitioned_graph::PG
  partitioned_vertices::Dictionary
  which_partition::Dictionary
end
```

then the constructors will accept Vectors or Dictionaries and be simplified to the level you showed above. I have done this now, so let me know what you think!

examples/partitioning.jl
```diff
@@ -0,0 +1,16 @@
...
+using Metis
+
+g = named_grid((4, 4))
+npartitions = 4
+
+pg_kahypar = PartitionedGraph(g; npartitions, backend="KaHyPar")
+pg_metis = PartitionedGraph(g; npartitions, backend="Metis")
+
+@show length(NamedGraphs.vertices(partitioned_graph(pg_kahypar))) ==
+  length(NamedGraphs.vertices(partitioned_graph(pg_metis))) ==
```
mtfishman, line 12: Can just use `nv(...)` from `Graphs.jl` instead of `length(vertices(...))`.

src/Graphs/partitionedgraphs/abstractpartitionedgraph.jl
```diff
@@ -0,0 +1,173 @@
...
+  return vertex_to_parent_vertex(unpartitioned_graph(pg), vertex)
+end
+edgetype(pg::AbstractPartitionedGraph) = edgetype(unpartitioned_graph(pg))
+function parent_graph_type(G::Type{<:AbstractPartitionedGraph})
+  return fieldtype(fieldtype(G, :graph), :parent_graph)
+end
+directed_graph(G::Type{<:AbstractPartitionedGraph}) = directed_graph(fieldtype(G, :graph))
+function undirected_graph(G::Type{<:AbstractPartitionedGraph})
+  return unddirected_graph(fieldtype(G, :graph))
+end
```
mtfishman, line 33: Any functions that refer directly to fields should one be defined for concrete subtypes like `PartitionedGraph`, they aren't appropriate as abstract definitions since we don't want to assume anything about what kind of fields a subtype might have.

I would try to organize these into a minimal set of functions which would be considered part of the interface, define them as `not_implemented()` for the abstract type, and then define some of these function in terms of those minimal functions.

Another suggestion is that for functions that return types, it's conventional (and I think clearer) to have them end with `_type`, for example calling it `undirected_graph_type(...)` instead of `undirected_graph(...)` if it returns a type as opposed to an instance.

src/Graphs/partitionedgraphs/partitionedgraph.jl
```diff
@@ -0,0 +1,137 @@
...
+
+function partition_edges(pg::PartitionedGraph, partition_edge::PartitionEdge)
+  psrc_vs, pdst_vs = vertices(pg, PartitionVertex(src(partition_edge))),
+  vertices(pg, PartitionVertex(dst(partition_edge)))
+  psrc_subgraph, pdst_subgraph, full_subgraph = subgraph(unpartitioned_graph(pg), psrc_vs),
+  subgraph(pg, pdst_vs),
+  subgraph(pg, vcat(psrc_vs, pdst_vs))
+
+  return setdiff(
+    NamedGraphs.edges(full_subgraph),
```
mtfishman, line 71: Is specifying `NamedGraphs.edges(...)` instead of just called `edges(...)` necessary?

src/Graphs/partitionedgraphs/partitionedgraph.jl
```diff
@@ -0,0 +1,137 @@
...
+  pg::PartitionedGraph, vertex, partition_vertex::PartitionVertex
+)
+  pv = parent(partition_vertex)
+  if pv ∉ keys(partitioned_vertices(pg))
+    insert!(partitioned_vertices(pg), pv, [vertex])
+  else
+    pg.partitioned_vertices[pv] = unique(vcat(vertices(pg, partition_vertex), [vertex]))
+  end
+
+  return insert!(NamedGraphs.which_partition(pg), vertex, pv)
```
mtfishman, line 96: `which_partition` instead of `NamedGraphs.which_partition`?

src/Graphs/partitionedgraphs/partitionedgraph.jl
```diff
@@ -0,0 +1,137 @@
...
+  end
+
+  return PartitionedGraph(sub_pg_graph, sub_partitioned_vertices), nothing
+end
+
+function induced_subgraph(
+  pg::PartitionedGraph, partition_verts::Vector{V}
+) where {V<:PartitionVertex}
+  return induced_subgraph(pg, vertices(pg, partition_verts))
+end
```
mtfishman, line 129: Seems like this could be an abstract definition, i.e. change to `AbstractPartitionedGraph` and `Vector{<:AbstractPartitionVertex}`.

src/Graphs/partitionedgraphs/partitionedgraph.jl
```diff
@@ -0,0 +1,137 @@
...
+
+function subgraph(pg::PartitionedGraph, partition_vertex::PartitionVertex)
+  return first(induced_subgraph(unpartitioned_graph(pg), vertices(pg, [partition_vertex])))
+end
+
+function induced_subgraph(pg::PartitionedGraph, partition_vertex::PartitionVertex)
+  return subgraph(pg, partition_vertex), nothing
+end
```
mtfishman, line 137: Seems like these could also be abstract definitions.

test/test_partitionedgraph.jl
```diff
@@ -0,0 +1,130 @@
...
+    named_comb_tree((4, 4)),
+    named_grid((2, 2, 2)),
+    NamedGraph(random_regular_graph(12, 3)),
+    NamedGraphs.triangular_lattice_graph(7, 7),
+  ]
+
+  for f in functions
+    for g in gs
+      pg = PartitionedGraph(g, [vertices(g)])
+      @test f(pg) == f(pg.graph)
```
mtfishman, line 123:
```suggestion
      @test f(pg) == f(unpartitioned_graph(pg))
```

test/test_partitionedgraph.jl
```diff
@@ -0,0 +1,130 @@
...
+  spanning_forest,
+  subvertices,
+  spanning_tree,
+  forest_cover,
+  rem_edge!,
+  PartitionEdge,
+  rem_edges!,
+  PartitionVertex,
+  parent,
+  _npartitions
```
mtfishman, line 13: Looks like this can be removed now.

src/Graphs/partitionedgraphs/partitionedgraph.jl
```diff
@@ -0,0 +1,137 @@
...
+    parent(which_partition(pg, src(edge))) => parent(which_partition(pg, dst(edge)))
+  )
+end
+
+function partition_edges(pg::PartitionedGraph, partition_edge::PartitionEdge)
+  psrc_vs, pdst_vs = vertices(pg, PartitionVertex(src(partition_edge))),
+  vertices(pg, PartitionVertex(dst(partition_edge)))
+  psrc_subgraph, pdst_subgraph, full_subgraph = subgraph(unpartitioned_graph(pg), psrc_vs),
+  subgraph(pg, pdst_vs),
+  subgraph(pg, vcat(psrc_vs, pdst_vs))
```
mtfishman, line 68: I would just separate these into separate lines, easier to read that way.

src/Graphs/partitionedgraphs/partitionedgraph.jl
```diff
@@ -0,0 +1,137 @@
...
+partitioned_graph(pg::PartitionedGraph) = getfield(pg, :partitioned_graph)
+unpartitioned_graph(pg::PartitionedGraph) = getfield(pg, :graph)
+partitioned_vertices(pg::PartitionedGraph) = getfield(pg, :partitioned_vertices)
+which_partition(pg::PartitionedGraph) = getfield(pg, :which_partition)
+function vertices(pg::PartitionedGraph, partition_vert::PartitionVertex)
+  return partitioned_vertices(pg)[parent(partition_vert)]
+end
+function vertices(
+  pg::PartitionedGraph, partition_verts::Vector{V}
+) where {V<:PartitionVertex}
```
mtfishman, line 50:
```suggestion
  pg::PartitionedGraph, partition_verts::Vector{<:PartitionVertex}
)
```
I only use `where` if it is really needed, for example if I am using a parameter inside the function or imposing some constraint between types.

src/Graphs/partitionedgraphs/partitionedge.jl
```diff
@@ -0,0 +1,8 @@
+struct PartitionEdge{V,E<:AbstractEdge{V}} <: AbstractPartitionEdge{V}
+  edge::E
+end
+
+parent(pe::PartitionEdge) = getfield(pe, :edge)
+src(pe::PartitionEdge) = PartitionVertex(src(parent(pe)))
+dst(pe::PartitionEdge) = PartitionVertex(dst(parent(pe)))
+PartitionEdge(p::Pair) = PartitionEdge(NamedEdge(p.first => p.second))
```
mtfishman, line 8:
```suggestion
PartitionEdge(p::Pair) = PartitionEdge(NamedEdge(first(p) => last(p)))
```
I generally try to avoid accessing fields directly.

src/Graphs/partitionedgraphs/abstractpartitionedgraph.jl
```diff
@@ -0,0 +1,173 @@
...
+  rem_vertex!(pg_new, partition_vertex)
+  return pg_new
+end
+
+function rem_vertices!(
+  pg::AbstractPartitionedGraph, partition_vertices::Vector{<:AbstractPartitionVertex}
+)
+  for pv in partition_vertices
+    rem_vertex!(pg, pv)
+  end
```
mtfishman, line 145:
```suggestion
  end
  return pg
```
Usually I return the object being modified as a convention.

src/Graphs/partitionedgraphs/abstractpartitionedgraph.jl
```diff
@@ -0,0 +1,173 @@
...
+)
+  return add_vertices!(pg, vertices, fill(partition_vertex, length(vertices)))
+end
+
+function rem_vertex!(pg::AbstractPartitionedGraph, vertex)
+  rem_vertex!(unpartitioned_graph(pg), vertex)
+  pv = which_partition(pg, vertex)
+  if length(vertices(pg, pv)) == 1
+    rem_vertex!(partitioned_graph(pg), parent(pv))
+  end
```
mtfishman, line 124: Maybe I'm missing the logic here, but shouldn't it be checking if the number of vertices in the partition is zero?

Also maybe we want a shorthand:
```julia
nv(pg::AbstractPartitionedGraph, pv::AbstractPartionedVertex) = length(vertices(pg, pv))
```

src/Graphs/partitionedgraphs/partitionedgraph.jl
```diff
@@ -0,0 +1,137 @@
...
+end
+
+function delete_from_vertex_map!(pg::PartitionedGraph, vertex)
+  pv = which_partition(pg, vertex)
+  vs = vertices(pg, pv)
+  delete!(partitioned_vertices(pg), parent(pv))
+  if length(vs) != 1
+    insert!(partitioned_vertices(pg), parent(pv), setdiff(vs, [vertex]))
+  end
+  return delete!(pg.which_partition, vertex)
```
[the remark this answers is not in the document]
mtfishman: Also I would return `pg` instead of whatever `delete!` outputs.

src/abstractnamedgraph.jl
mtfishman: I think what makes sense with this is defining generic `set_vertex!`/`unset_vertex!` and `insert_vertex!`/`delete_vertex!` for general `AbstractGraph`, written in terms of `add_vertex!`/`rem_vertex!`. Then for `AbstractNamedGraph` we just define `add_vertex!`/`rem_vertex!`, and the rest would be defined automatically. What do you think?

Even if we stick to just defining them for `AbstractNamedGraph` for now, I think we should write `set_vertex!`/`unset_vertex!`/`insert_vertex!`/`delete_vertex!` in terms of `add_vertex!`/`rem_vertex!`, since those should be considered the "core" required interface functions, to follow how it works in `Graphs.jl`.

src/Graphs/partitionedgraphs/partitionedgraph.jl
mtfishman: The new design looks good, thanks!

src/Graphs/partitionedgraphs/abstractpartitionedgraph.jl
mtfishman: Oh I see, I guess it counts the vertices in the partition based on `partitioned_vertices(pg)`, which only get updated in `delete_from_vertex_map!` in the line below, so at this point it still looks like `pg` would have one vertex instead of zero in the case where a partition will end up empty.

Maybe it would be clearer to write it as:
```julia
function rem_vertex!(pg::AbstractPartitionedGraph, vertex)
  pv = which_partition(pg, vertex)
  delete_from_vertex_map!(pg, pv, vertex)
  rem_vertex!(unpartitioned_graph(pg), vertex)
  if iszero(nv(pg, pv))
    rem_vertex!(partitioned_graph(pg), parent(pv))
  end
  return pg
end
```

This also assumes we refactor `delete_from_vertex_map!` slightly:
```julia
function delete_from_vertex_map!(pg::PartitionedGraph, vertex)
  pv = which_partition(pg, vertex)
  return delete_from_vertex_map!(pg, pv, vertex)
end

function delete_from_vertex_map!(pg::PartitionedGraph, partitioned_vertex::PartitionedVertex, vertex)
  # Implementation
  return pv
end
```

src/Graphs/partitionedgraphs/abstractpartitionedgraph.jl
JoeyT1994: Yeah that makes sense. I think `undirected_graph(....)` in this instance though is just returning `true` or `false` depending on the type?

JoeyT1994: Yeah that makes sense! Although that should probably change on lines 39 and 40 of `abstractnamedgraphs.jl` too then?  As I think that is missing the `_type` underscore

mtfishman: Yeah, if functions elsewhere aren't following that convention they should be changed as well.

src/Graphs/partitionedgraphs/abstractpartitionedgraph.jl
JoeyT1994: This comes though down to whether `nv(pg,pv)` determines the number of vertices from the `partitioned_graph` itself or the `vertex_map` ? I changed it to check it based on the `vertex_map` instead of the `partitioned_graph`  so we can use `iszero` but it is not clear to me which is best.

src/Graphs/partitionedgraphs/partitionedgraph.jl
```diff
@@ -0,0 +1,143 @@
...
+unpartitioned_graph(pg::PartitionedGraph) = getfield(pg, :graph)
+partitioned_vertices(pg::PartitionedGraph) = getfield(pg, :partitioned_vertices)
+which_partition(pg::PartitionedGraph) = getfield(pg, :which_partition)
+parent_graph_type(PG::Type{<:PartitionedGraph}) = fieldtype(PG, :graph)
+function directed_graph_type(PG::Type{<:PartitionedGraph})
+  return directed_graph_type(fieldtype(PG, :graph))
+end
+function undirected_graph_type(PG::Type{<:PartitionedGraph})
+  return undirected_graph_type(fieldtype(PG, :graph))
+end
```
JoeyT1994, line 51: Currently functions like this just operate directly on the `unpartitioned_graph` and return a `NamedDiGraph` or `NamedGraph` type.
This gets functions like `spanning_tree(pg)` to work (which outputs a `NamedGraph` also and just operates on `unpartitioned_graph`) but maybe is a bit strange because I would imagine they should all ideally return a `PartitionedGraph`.  For instance `directed_graph(pg)` should probably convert both the `unpartitioned_graph` and `partitioned_graph` to a `NamedDiGraph`? 
So, do we:
1) Leave as is
2) Modify these functions to return `PartitionedGraph`s, which would require new functions like `bfs_tree(pg::PartitionedGraph)` 
3) Take away that functionality and have functions like `spanning_tree(pg)`, `directed_graph(pg)` not work for now (the user has to either directly do `spanning_tree(partitioned_graph(pg))` or `spanning_tree(unpartitioned_graph(pg))`. We can then add in `spanning_tree(pg)`, `directed_graph(pg)` etc on the concrete types we desire later on

src/Graphs/partitionedgraphs/abstractpartitionedgraph.jl
mtfishman: Isn't the information for the number of vertices in a partition only stored in the `vertex_map`? How would you get it from the `partitioned_graph`?

src/Graphs/partitionedgraphs/partitionedgraph.jl
mtfishman: Good point. I agree with your assessment, `directed_graph_type`, `undirected_graph_type`, `spanning_tree`, `bfs_tree`, etc. of `PartitionedGraph` should return `PartitionedGraph`. I think option 3. is good for now, we can always add the definitions that work on `PartitionedGraph` later.

src/Graphs/partitionedgraphs/abstractpartitionedgraph.jl
JoeyT1994: Yeah this is true. Not sure what I meant when saying that: you can only determine whether the number of vertices is zero or nonzero from the `partitioned_graph`.

I have made the code in line with your suggestion above. To do this we need to avoid `nv(pg,pv)` erroring when `pv` is not in the `vertex_map` of `partitioned_graph(pg)` (for instance because it just got deleted from the call to `delete_from_vertex_map!`). I made `vertices(pg,pv)` return an empty list to handle this case.

src/Graphs/partitionedgraphs/partitionedgraph.jl
JoeyT1994: Okay let me work option 3 in tomorrow morning.

src/Graphs/partitionedgraphs/abstractpartitionedgraph.jl
mtfishman: Hmm let's consider what `vertices` should return if the partition doesn't exist... I would lean towards being conservative about that and throw an error, maybe you can add a check `has_vertex(pg, pv)` in cases where you think `pg` may not have the partition? The part that mostly concerns me about returning an empty list is then a partition that doesn't exist acts the same as one that does exist but doesn't have any vertices, which doesn't feel right to me.

JoeyT1994: Okay yeah I can add that check. I have made it check `haskey(partitioned_vertices(pg), pv)` directly? Because it also raises the question what should `has_vertex(pg, pv)` check for `pv` in? The `partitioned_graph` or the `vertex_map` or both?

mtfishman: I think it should just check whichever one is faster to check, I don't think it really matters.

mtfishman: I.e. shouldn't `partitioned_graph` and `vertex_map` be consistent about whether or not `pv` exists in the partitioned graph? I think the answer is yes, so you may as well just forward to `partitioned_graph`, since that seems like the cleanest and most general approach (that design will work for various `AbstractPartitionedGraph` subtypes, many functions accepting `PartitionedVertex` can be implemented by just forwarding to `partitioned_graph` so we should bias towards that code style, etc.).

JoeyT1994: Okay that's fair enough. I have changed it to do the forwarded version.

mtfishman:
@JoeyT1994 could you remind me of the status of this? Is the only thing remaining to change the code organization of `insert_vertex!` and related functions?

src/abstractnamedgraph.jl
JoeyT1994: Okay I made `add_vertex!`/`rem_vertex!` the core required interface function for an `AbstractGraph` and then `insert_vertex!`/`delete_vertex!` is written in terms of that.

JoeyT1994:
Yeah I believe that is the case! And I just re-organised it so that `add_vertex!` is required for the core interface of an `abstractgraph`

examples/partitioning.jl
```diff
@@ -0,0 +1,14 @@
...
+g = named_grid((4, 4))
+npartitions = 4
+
+pg_kahypar = PartitionedGraph(g; npartitions, backend="KaHyPar")
+pg_metis = PartitionedGraph(g; npartitions, backend="Metis")
+
+@show nv(partitioned_graph(pg_kahypar)) == nv(partitioned_graph(pg_metis)) == npartitions
+
+@show pg_kahypar isa PartitionedGraph
+@show pg_metis isa PartitionedGraph
```
mtfishman, line 14: Can this file be included in the tests if it isn't already?

src/Graphs/abstractgraph.jl
```diff
@@ -384,6 +386,31 @@ function mincut_partitions(graph::AbstractGraph, distmx=weights(graph))
   return parts[1], parts[2]
 end
 
+function insert_vertex!(graph::AbstractGraph, vertex)
+  z = add_vertex!(graph, vertex)
```
mtfishman, line 390: Let's call it `in_graph` or something more descriptive like that instead of `z`. Same for `delete_vertex!`.

src/Graphs/abstractgraph.jl
```diff
@@ -384,6 +386,31 @@ function mincut_partitions(graph::AbstractGraph, distmx=weights(graph))
...
+  if !z
+    error("Duplicate vertices are not allowed")
+  else
+    return graph
+  end
```
mtfishman, line 395:
```suggestion
end
return graph
```
I've started to prefer this style, it makes it clearer what is being returned (instead of hiding the return inside a clause of the if-statement). Same for `delete_vertex!`.

examples/partitioning.jl
JoeyT1994: It currently isn't because I didn't want to have the library require `KaHyPar` and `Metis` as a dependency (as there are only going to be used if the user wants to call them for graph partitioning).
Is there any way to have the test only run if they are in the dependencies?

mtfishman: We can require `KaHyPar` and/or `Metis` in the tests but not for the core library itself (i.e. it's fine to have test dependencies).

src/abstractnamedgraph.jl
```diff
@@ -381,44 +381,42 @@ function union(graph1::AbstractNamedGraph, graph2::AbstractNamedGraph)
...
+
+    # Insert the last vertex into the position of the vertex
+    # that is being deleted, then remove the last vertex.
+    # TODO: Make this more generic
+    graph.vertex_to_parent_vertex[last_vertex] = parent_vertex
+    # TODO: Make this more generic
+    delete!(graph.vertex_to_parent_vertex, vertex)
+
+    return true
+  end
```
mtfishman, line 405:
```suggestion
  end
    parent_vertex = vertex_to_parent_vertex(graph, vertex)
    rem_vertex!(parent_graph(graph), parent_vertex)

    # Insert the last vertex into the position of the vertex
    # that is being deleted, then remove the last vertex.
    last_vertex = last(vertices(graph))
    vertices(graph)[parent_vertex] = last_vertex
    last_vertex = pop!(vertices(graph))

    # Insert the last vertex into the position of the vertex
    # that is being deleted, then remove the last vertex.
    # TODO: Make this more generic
    graph.vertex_to_parent_vertex[last_vertex] = parent_vertex
    # TODO: Make this more generic
    delete!(graph.vertex_to_parent_vertex, vertex)

    return true
```
Same kind of change elsewhere, like `add_vertex!`. This helps clarify the code logic, to separate out the code into "escape hatches" and then the actual implementation. Also I think it is easier to read code that isn't formatted into an if-statement.

examples/partitioning.jl
JoeyT1994: Ah yes I see how to do this now. Thanks!

src/Graphs/partitionedgraphs/abstractpartitionedgraph.jl
```diff
@@ -0,0 +1,193 @@
...
+end
+
+#Vertex addition and removal. I think it's important not to allow addition of a vertex without specification of PV
+function add_vertex!(
+  pg::AbstractPartitionedGraph, vertex, partition_vertex::AbstractPartitionVertex
+)
+  add_vertex!(unpartitioned_graph(pg), vertex)
+  add_vertex!(partitioned_graph(pg), parent(partition_vertex))
+
+  return insert_to_vertex_map!(pg, vertex, partition_vertex)
```
mtfishman, line 104:
```suggestion
  insert_to_vertex_map!(pg, vertex, partition_vertex)
  return pg
```

src/Graphs/partitionedgraphs/abstractpartitionedgraph.jl
```diff
@@ -0,0 +1,193 @@
...
+    add_vertex!(pg, v, pv)
+  end
+
+  return pg
+end
+
+function add_vertices!(
+  pg::AbstractPartitionedGraph, vertices::Vector, partition_vertex::AbstractPartitionVertex
+)
+  return add_vertices!(pg, vertices, fill(partition_vertex, length(vertices)))
```
mtfishman, line 123:
```suggestion
  add_vertices!(pg, vertices, fill(partition_vertex, length(vertices)))
  return pg
```

src/Graphs/partitionedgraphs/abstractpartitionedgraph.jl
```diff
@@ -0,0 +1,193 @@
...
+  if !haskey(partitioned_vertices(pg), parent(pv))
+    rem_vertex!(partitioned_graph(pg), parent(pv))
+  end
+  return pg
+end
+
+function rem_vertex!(
+  pg::AbstractPartitionedGraph, partition_vertex::AbstractPartitionVertex
+)
+  return rem_vertices!(pg, vertices(pg, partition_vertex))
```
mtfishman, line 139:
```suggestion
  rem_vertices!(pg, vertices(pg, partition_vertex))
  return pg
```

src/Graphs/partitionedgraphs/abstractpartitionedgraph.jl
```diff
@@ -0,0 +1,193 @@
...
+  return pg
+end
+
+function rem_vertices(
+  pg::AbstractPartitionedGraph, partition_vertices::Vector{<:AbstractPartitionVertex}
+)
+  pg_new = copy(rem_partition_vertex)
+  rem_vertices!(pg_new, partition_vertices)
+  return pg_new
+end
```
mtfishman, line 164: Isn't this already covered by https://github.com/mtfishman/NamedGraphs.jl/blob/1b356a9cd31dca18e1762fdff30989bd042eebd2/src/Graphs/abstractgraph.jl#L424-L428?

src/Graphs/abstractgraph.jl
```diff
@@ -384,6 +386,31 @@ function mincut_partitions(graph::AbstractGraph, distmx=weights(graph))
   return parts[1], parts[2]
 end
 
+function insert_vertex!(graph::AbstractGraph, vertex)
+  in_graph = !add_vertex!(graph, vertex)
+  if in_graph
+    error("Duplicate vertices are not allowed")
+  end
+
```
mtfishman, line 394:
```suggestion
```
I see a few other places with extraneous empty lines as well.

mtfishman:
I think after addressing the last few comments I left this should be good to go.

src/Graphs/partitionedgraphs/abstractpartitionedgraph.jl
JoeyT1994: Yes good spot! It is redundant so I have removed it.

src/Graphs/abstractgraph.jl
JoeyT1994: Ah I thought the formatter dealt with that. Apparently not.

JoeyT1994:
Great! I believe I have resolved the above.

src/Graphs/abstractgraph.jl
```diff
@@ -1,5 +1,7 @@
 directed_graph(::Type{<:AbstractGraph}) = error("Not implemented")
 undirected_graph(::Type{<:AbstractGraph}) = error("Not implemented")
+add_vertex!(graph::AbstractGraph, vertex) = error("Not implemented")
+rem_vertex!(graph::AbstractGraph, vertex) = error("Not implemented")
```
mtfishman, line 4: Hmm I don't think we want this, seems like [type piracy](https://docs.julialang.org/en/v1/manual/style-guide/#Avoid-type-piracy) (i.e. overwriting methods from `Graphs.jl` with types defined in `Graphs.jl`).

test/test_partitionedgraph.jl
```diff
@@ -0,0 +1,113 @@
...
+
+  partitions = [[(i, j) for j in 1:ny] for i in 1:nx]
+  pg = PartitionedGraph(g, partitions)
+
+  pv = PartitionVertex(5)
+  v_set = vertices(pg, pv)
+  edges_involving_v_set = boundary_edges(g, v_set)
+
+  #Strip the middle column from pg via the partitioned graph vertex, and make a new pg
+  NamedGraphs.rem_vertex!(pg, pv)
```
mtfishman, line 48:
```suggestion
  rem_vertex!(pg, pv)
```
and similar elsewhere.

PR #46, opened 2024-01-23
Fix bug in `PartitionEdge` with mixed vertex types, generalize `partitioned_vertices` to `AbstractGraph`

JoeyT1994:
This is a quick PR to fix a bug and make a few small minor changes:

1.  A bug was fixed in the definition of `PartitionEdge` which was preventing the construction of a `PartitionEdge` from a `NamedEdge{Tuple{V,V'}}` where `V`  and `V'` are different types
2. `partitioned_vertices(g; args...)` which passes to various partitioning backends now acts on an `AbstractGraph` type so one can directly pass types such as a `DataGraph` or `ITensorNetwork` to it.


mtfishman:
I was only suggesting having `subgraph(::PartitionedGraph, ::PartitionVertex)` return an unpartitioned graph, I still think `subgraph(::PartitionedGraph, ::Vector{<:PartitionVertex})` should return a partitioned graph.

JoeyT1994:
I see. I actually think that is already the behavior so I have reverted it back to that. I just forgot to call it that way in `ITensorNetwork.jl`

mtfishman:
Oh yeah, I see that defined here: https://github.com/mtfishman/NamedGraphs.jl/blob/v0.1.16/src/Graphs/partitionedgraphs/abstractpartitionedgraph.jl#L168-L176

PR #47, opened 2024-01-25
Allow partitionedge and partitionedges to accept pairs

JoeyT1994:
A quick PR which allows `partitionedge(pg, edge)` and `partitionedges(pg, edges)` to accept a `Pair` or `Vector{Pair}` alongside `AbstractEdge` and `Vector{AbstractEdge}` types.


src/Graphs/partitionedgraphs/partitionedgraph.jl
```diff
@@ -67,6 +67,9 @@ function partitionedge(pg::PartitionedGraph, edge::AbstractEdge)
   )
 end
 
+partitionedge(pg::PartitionedGraph, p::Pair) = partitionedge(pg, NamedEdge(p))
+partitionedges(pg::PartitionedGraph, ps::Vector{Pair}) = partitionedges(pg, NamedEdge.(ps))
+
 #Lets filter out any self-edges from this. Although this makes it a bit consistent with partitionedge
 function partitionedges(pg::PartitionedGraph, edges::Vector{<:AbstractEdge})
```
mtfishman, line 74:
```suggestion
function partitionedges(pg::PartitionedGraph, edges::Vector)
```
should mean you don't need a separate definition for `Vector{<:Pair}`.

src/Graphs/partitionedgraphs/partitionedgraph.jl
```diff
@@ -67,6 +67,9 @@ function partitionedge(pg::PartitionedGraph, edge::AbstractEdge)
   )
 end
 
+partitionedge(pg::PartitionedGraph, p::Pair) = partitionedge(pg, NamedEdge(p))
```
mtfishman, line 70:
```suggestion
partitionedge(pg::PartitionedGraph, p::Pair) = partitionedge(pg, edgetype(pg)(p))
```

mtfishman:
Thanks, could you add some tests?

src/Graphs/partitionedgraphs/partitionedgraph.jl
JoeyT1994: The definition of `partitionedge(pg, edges::Vector{<:AbstractEdge})` requires it to call `src` and `dst` on the members of `edges`? To avoid returning self-edges. These aren't defined on a `Pair`. Not sure how quite how to handle this part tbf. Do we drop the check for self-edges? Do we put it into `partitionedge` too?

mtfishman: Ah I didn't see that it was using `src` and `dst`, I see that makes the function not handle both `Pair` and `AbstractEdge` automatically.

Something we could do is define `is_self_loop` (inspired by the function [Graphs.has_self_loops](https://juliagraphs.org/Graphs.jl/v1.9/core_functions/core/#Graphs.has_self_loops-Tuple{AbstractGraph})):
```julia
is_self_loop(e::AbstractEdge) = src(e) == dst(e)
is_self_loop(e::Pair) = first(e) == last(e)

function partitionedges(pg::PartitionedGraph, edges::Vector)
  return filter(is_self_loop, unique([partitionedge(pg, e) for e in edges]))
end
```

JoeyT1994:
Sure I added a few tests.

src/Graphs/partitionedgraphs/partitionedgraph.jl
JoeyT1994: Okay that sounds reasonable, I have added that.

PR #50, opened 2024-02-15
Fix bug in `copy(pg::PartitionedGraph)`

JoeyT1994:
Quick fix of a bug in `copy(pg::PartitionedGraph)` where the wrong field name was being used.
A test has been added for `copy(pg)` so the bug would be caught if it happens again.


mtfishman:
Thanks!

PR #51, opened 2024-02-20
`boundary_partitionedges`

JoeyT1994:
New function `boundary_partitionedges(pg::PartitionedGraph, partitionvertices::Vector{<:PartitionVertex})` which returns all the `partitionedges` (in `PartitionEdge` format) incoming to the region specified by `partitionvertices`. Tests are also added.


src/Graphs/partitionedgraphs/partitionedgraph.jl
```diff
@@ -94,12 +94,26 @@ function edges(pg::PartitionedGraph, partitionedges::Vector{<:PartitionEdge})
...
+)
+  return PartitionEdge.(
+    boundary_edges(partitioned_graph(pg), parent.(partitionvertices); kwargs...)
+  )
+end
+
+function boundary_partitionedges(
+  pg::PartitionedGraph, partitionvertex::PartitionVertex; kwargs...
+)
+  return boundary_partitionedges(pg, PartitionVertex[partitionvertex]; kwargs...)
```
mtfishman, line 108:
```suggestion
  return boundary_partitionedges(pg, [partitionvertex]; kwargs...)
```
Julia should already infer the type is `Vector{<:PartitionVertex}`.

src/NamedGraphs.jl
```diff
@@ -187,6 +187,7 @@ export NamedGraph,
   partitionvertex,
   partitionvertices,
   partitioned_vertices,
+  boundary_partitionedges,
```
mtfishman, line 190: Let's avoid exporting for now and reassess exports going forward.

For example, ideally we would have a `PartitionedGraphs` module, which then exports partition-related functions. So better off not exporting so we don't have to backtrack later when we move to that design.

mtfishman:
Could you add a test for `copy(::PartitionedGraph)`?

JoeyT1994:
Sure I just added that.

mtfishman:
Thanks!

JoeyT1994:
Can we move to a new version of NamedGraphs following this PR? Just so the tests for the ITensorNetworks PR: https://github.com/mtfishman/ITensorNetworks.jl/pull/139 pass

mtfishman:
Thanks for the reminder, registering it now.

PR #77, opened 2024-05-14
Add vertices_at_distance and next_nearest_neighbors

JoeyT1994:
PR which moves the `nth_nearest_neighbors` functionality from `ITensorNetworks.jl` to the `GraphsExtensions` module in this library. 
A few short tests are added, although maybe more of the `GraphsExtensions` code should be tested in the future (leaving for another PR).


mtfishman:
We should assess the name of this function.

Apparently the terminology "k nearest neighbors" has a different meaning in the graph community. For example, see the function [k_nearest_neighbors](https://networkx.org/documentation/networkx-1.11/reference/generated/networkx.algorithms.assortativity.k_nearest_neighbors.html) in older versions of NetworkX, which was renamed to [average_degree_connectivity](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.assortativity.average_degree_connectivity.html) in NetworkX v3.

The closest function in NetworkX that I can find to the functionality we want here is [descendants_at_distance](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.traversal.breadth_first_search.descendants_at_distance.html), though the terminology "descendants" makes it seem like it is for directed graphs.

Perhaps we could call it `vertices_at_distance`. Additionally, it may be better to implement it using [Graphs.neighborhood_dists](https://juliagraphs.org/Graphs.jl/stable/algorithms/connectivity/#Graphs.neighborhood_dists-Union{Tuple{T},%20Tuple{U},%20Tuple{AbstractGraph{T},%20Integer,%20Any},%20Tuple{AbstractGraph{T},%20Integer,%20Any,%20AbstractMatrix{U}}}%20where%20{U%3C:Real,%20T%3C:Integer}). In that way only a single neighborhood needs to be computed, and then we can filter the vertices at a certain distance, possibly using [SplitApplyCombine.jl](https://github.com/JuliaData/SplitApplyCombine.jl).

JoeyT1994:
`vertices_at_distance` seems like a reasonable name to me. I guess `k_nearest_neighbors` is using `k` to mean degree but that is not a very transparent name.

Okay yeah it should be easier to write using `neighborhood_dists` rather than recursively calling the function. Let me change that.

src/lib/GraphsExtensions/src/neighbors.jl
```diff
@@ -1,8 +1,10 @@
-using Graphs: AbstractGraph, neighborhood
+using Graphs: AbstractGraph, neighborhood_dists
 
-function nth_nearest_neighbors(g::AbstractGraph, v, n::Int)
-  isone(n) && return neighborhood(g, v, 1)
-  return setdiff(neighborhood(g, v, n), neighborhood(g, v, n - 1))
+function vertices_at_distance(g::AbstractGraph, vertex, n::Int)
+  neighborhood = [first(v) for v in neighborhood_dists(g, vertex, n)]
+  closer_neighborhood = [first(v) for v in neighborhood_dists(g, vertex, n - 1)]
```
mtfishman, line 5: I believe we can achieve what we want with only a single call to `neighborhood_dists`, since `neighborhood_dists(g, vertex, n)` contains the vertices and their distances from `vertex` so we should be able to filter out the ones that are at a distance `n` from `vertex`.

Also let's rename `n` to `distance` to make the code clearer.

mtfishman: For example:
```julia
julia> using Graphs

julia> g = path_graph(9)
{9, 8} undirected simple Int64 graph

julia> neighborhood_dists(g, 5, 3)
7-element Vector{Tuple{Int64, Int64}}:
 (5, 0)
 (4, 1)
 (6, 1)
 (3, 2)
 (7, 2)
 (2, 3)
 (8, 3)

julia> filter(==(3) ∘ last, neighborhood_dists(g, 5, 3))
2-element Vector{Tuple{Int64, Int64}}:
 (2, 3)
 (8, 3)

julia> map(first, filter(==(3) ∘ last, neighborhood_dists(g, 5, 3)))
2-element Vector{Int64}:
 2
 8
```

test/test_graphsextensions.jl
mtfishman, line 1: The rest of the tests for `GraphsExtensions` are here: https://github.com/mtfishman/NamedGraphs.jl/blob/v0.6.0/src/lib/GraphsExtensions/test/runtests.jl

JoeyT1994: Thanks! I moved the test for `vertices_at_distance` to there.

src/lib/GraphsExtensions/src/neighbors.jl
```diff
@@ -1,10 +1,8 @@
 using Graphs: AbstractGraph, neighborhood_dists
 
-function vertices_at_distance(g::AbstractGraph, vertex, n::Int)
-  neighborhood = [first(v) for v in neighborhood_dists(g, vertex, n)]
-  closer_neighborhood = [first(v) for v in neighborhood_dists(g, vertex, n - 1)]
-  iszero(n) && return neighborhood
-  return setdiff(neighborhood, closer_neighborhood)
+function vertices_at_distance(g::AbstractGraph, vertex, distance::Int)
+  n_dists = neighborhood_dists(g, vertex, distance)
+  return map(first, filter(==(distance) ∘ last, n_dists))
```
mtfishman, line 5:
```suggestion
  vertices_and_distances = neighborhood_dists(g, vertex, distance)
  return map(first, filter(==(distance) ∘ last, vertices_and_distances))
```

mtfishman:
Thanks!

PR #82, opened 2024-05-28
Remove degree-zero vertices from steiner_tree output

### LINKED ISSUE #80: [BUG] Possible bug in `steiner_tree`
Currently, the `steiner_tree` function reproduces the same behavior that it has for `SimpleGraph` where in that case, all vertices in the range `1:nv(steiner_tree(g,...))` remain. However, it may be that NamedGraphs should not have this behavior, and that only vertices which are reachable by traversing the Steiner tree should be included in the returned graph.

Here is a small code demonstrating the current behavior and output:
```
using Graphs: add_edge!, edges, nv, steiner_tree, vertices
using NamedGraphs

g = NamedGraph(["a","b","c"])
add_edge!(g,"a","c")
add_edge!(g,"a","b")

st = steiner_tree(g,["a","c"])
@show st
```

Output:
```
st = NamedGraph{String} with 3 vertices:
3-element NamedGraphs.OrderedDictionaries.OrderedIndices{String}
 "a"
 "b"
 "c"

and 1 edge(s):
"a" => "c"
```

and one can confirm that in this case `nv(st) == 3`.

emstoudenmire:
This PR fixes #80 by removing any vertices of degree zero from the NamedGraph output by the function `steiner_tree`. It also adds a test of this new behavior. (The graph already used in the previous test resulting in many degree-zero vertices so was already a good test case.)

### DIFF
--- a/src/steiner_tree.jl
+++ b/src/steiner_tree.jl
@@ -9,13 +9,9 @@ using SimpleTraits: SimpleTraits, Not, @traitfn
     map(v -> vertex_positions(g)[v], term_vert),
     dist_matrix_to_position_dist_matrix(g, distmx),
   )
-  named_st = typeof(g)(
-    position_tree, map(v -> ordered_vertices(g)[v], vertices(position_tree))
-  )
-  # Detect and remove vertices of degree zero
-  zero_verts = filter(v -> degree(named_st, v) == 0, vertices(named_st))
-  for v in zero_verts
-    rem_vertex!(named_st, v)
+  tree = typeof(g)(position_tree, map(v -> ordered_vertices(g)[v], vertices(position_tree)))
+  for v in copy(vertices(tree))
+    iszero(degree(tree, v)) && rem_vertex!(tree, v)
   end
-  return named_st
+  return tree
 end
--- a/src/steiner_tree.jl
+++ b/src/steiner_tree.jl
@@ -9,5 +9,13 @@ using SimpleTraits: SimpleTraits, Not, @traitfn
     map(v -> vertex_positions(g)[v], term_vert),
     dist_matrix_to_position_dist_matrix(g, distmx),
   )
-  return typeof(g)(position_tree, map(v -> ordered_vertices(g)[v], vertices(position_tree)))
+  named_st = typeof(g)(
+    position_tree, map(v -> ordered_vertices(g)[v], vertices(position_tree))
+  )
+  # Detect and remove vertices of degree zero
+  zero_verts = filter(v -> degree(named_st, v) == 0, vertices(named_st))
+  for v in zero_verts
+    rem_vertex!(named_st, v)
+  end
+  return named_st
 end
--- a/test/test_namedgraph.jl
+++ b/test/test_namedgraph.jl
@@ -675,7 +675,10 @@ end
     st = steiner_tree(g, terminal_vertices)
     es = [(1, 2) => (1, 3), (1, 3) => (1, 4), (1, 4) => (2, 4), (2, 4) => (3, 4)]
     @test ne(st) == 4
-    @test nv(st) == 12
+    @test nv(st) == 5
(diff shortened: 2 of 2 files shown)


src/steiner_tree.jl
```diff
@@ -9,5 +9,13 @@ using SimpleTraits: SimpleTraits, Not, @traitfn
     map(v -> vertex_positions(g)[v], term_vert),
     dist_matrix_to_position_dist_matrix(g, distmx),
   )
-  return typeof(g)(position_tree, map(v -> ordered_vertices(g)[v], vertices(position_tree)))
+  named_st = typeof(g)(
```
mtfishman, line 12:
```suggestion
  tree = typeof(g)(
```
I've been using the convention that I call graphs `graph` or `g`, trees `tree` or `t`, etc. in this package, with the presumption that they are all named or generic and then underlying simple/integer vertex graphs are `position_graph`, `position_tree`, etc. (meaning they are graphs where the vertices are the integer positions of the vertices of the corresponding named graphs in the ordered list of named vertices).

src/steiner_tree.jl
```diff
@@ -9,5 +9,13 @@ using SimpleTraits: SimpleTraits, Not, @traitfn
...
+    position_tree, map(v -> ordered_vertices(g)[v], vertices(position_tree))
+  )
+  # Detect and remove vertices of degree zero
+  zero_verts = filter(v -> degree(named_st, v) == 0, vertices(named_st))
+  for v in zero_verts
+    rem_vertex!(named_st, v)
+  end
```
mtfishman, line 19:
```suggestion
  for v in vertices(tree)
    iszero(degree(tree, v)) && rem_vertex!(tree, v)
  end
```
Avoids the intermediate data structure and a bit more readable.

src/steiner_tree.jl
emstoudenmire: Makes sense – I'll change that

src/steiner_tree.jl
emstoudenmire: I was worried that mixing the two steps this way would be dangerous, since I originally tried a similar pattern on a `SimpleGraph` and it failed badly since vertex labels kept getting rewritten.

But here I suppose it's totally safe since the vertex names are unique.

mtfishman: I see, I don't anticipate an issue like that here since as you say the vertex names are unique.

emstoudenmire: Unfortunately the above version does lead to a bug. The reason is that `vertices(tree)` returns a reference, not a copy, to the vertices, so as `rem_vertex!` acts, it reduces the value of `nv(tree)` on the fly and the iteration ends too early. One fix that I tested could be:
```
  verts = copy(vertices(tree))
  for v in verts
    iszero(degree(tree, v)) && rem_vertex!(tree,v)
  end
```

emstoudenmire: Also this works:
```
  for v in copy(vertices(tree))
    iszero(degree(tree, v)) && rem_vertex!(tree,v)
  end
```

mtfishman: I was wondering about that, that sounds like a good fix.

test/test_namedgraph.jl
```diff
@@ -675,7 +675,8 @@ end
     st = steiner_tree(g, terminal_vertices)
     es = [(1, 2) => (1, 3), (1, 3) => (1, 4), (1, 4) => (2, 4), (2, 4) => (3, 4)]
     @test ne(st) == 4
-    @test nv(st) == 12
+    @test nv(st) == 5
+    @test !any(v -> degree(st, v) == 0, vertices(st))
```
mtfishman, line 679:
```suggestion
    @test !any(v -> iszero(degree(st, v)), vertices(st))
```
I like using the `iszero` function for zero comparison.

PR #91, opened 2025-03-27
Add Cycle Counting and EdgeSubgraphs

JoeyT1994:
This PR achieves several things related to cycle counting (https://en.wikipedia.org/wiki/Cycle_(graph_theory)) and edge-induced subgraphs.

1. The `simplecycles_limited_length` (https://juliagraphs.org/Graphs.jl/stable/algorithms/cycles) function from `Graphs.jl` is extended to work for the `AbstractNamedGraph` type. It counts all cycles in the graph up to a specified length.
2. The `unique_simplecycles_limited_length` function is then added which removes cycles of length 2 (which aren't actually cycles, I will raise an Issue with this in `Graphs.jl`) and also treats cycles that are just rotations of each other as identical.
3. The `edgeinduced_subgraph` function is added for getting an edge-induced subgraph of an `AbstractGraph`.

This then allows introduction of the algorithm: `edgeinduced_subgraphs_no_leaves(g::AbstractGraph, max_number_of_edges::Int64)` which finds all connected edgeinduced_subgraphs  with up to `max_number_of_edges` edges and without any vertices with `degree == 1`. These graphs are the non-zero connected terms that appear in belief-propagation based loop correction algorithms (see https://arxiv.org/abs/2409.03108 and  https://arxiv.org/abs/2503.05693).


src/simplecycles.jl
```diff
@@ -0,0 +1,57 @@
+using Combinatorics: powerset
+using Graphs: Graphs, is_tree, simplecycles_limited_length
+using NamedGraphs.GraphsExtensions: add_edges, disjoint_union, edge_subgraph
+using SplitApplyCombine: group
+
+function Graphs.simplecycles_limited_length(g::AbstractNamedGraph, max_cycle_size::Int64)
+  vs = collect(vertices(g))
+  cycles = simplecycles_limited_length(position_graph(g), max_cycle_size)
+  cycles = [[vs[i] for i in cycle] for cycle in cycles]
```
mtfishman, line 9:
```suggestion
  cycles = map(cycle -> map(i -> vs[i], cycle), cycles)
```
Mostly a style thing, but I prefer using `map` (it is a bit more generic since in general it can preserve the type of the input collection).

src/simplecycles.jl
```diff
@@ -0,0 +1,57 @@
...
+  cycles = first.(collect(values(cycles)))
+
+  return cycles
+end
+
+function cycle_to_path(g::AbstractNamedGraph, cycle::Vector)
+  es = [NamedEdge(cycle[i] => cycle[i + 1]) for i in 1:(length(cycle) - 1)]
+  final_edge = NamedEdge(first(cycle) => last(cycle))
+  return vcat(es, final_edge)
+end
```
mtfishman, line 28: I'm confused by this function, it doesn't look like the input graph `g` is used.

src/simplecycles.jl
```diff
@@ -0,0 +1,57 @@
...
+
+function unique_cyclesubgraphs_limited_length(g::AbstractNamedGraph, max_cycle_size::Int64)
+  cycles = unique_simplecycles_limited_length(g, max_cycle_size)
+  paths = cycle_to_path.((g,), cycles)
+  return edge_subgraph.((g,), paths)
+end
+
+"""
+Enumerate all unqiue, connected edgesubgraphs without any leaf vertices (degree 1) and with Nedges <= max_number_of_edges
```
mtfishman, line 37:
```suggestion
    edgeinduced_subgraphs_no_leaves(g::AbstractNamedGraph, max_number_of_edges::Int)
    
Enumerate all unique, connected edgesubgraphs without any leaf vertices (degree 1) and with Nedges <= max_number_of_edges
```
Docstrings generally start with the function signature formatted in the way I show above. Also I fixed a typo.

src/simplecycles.jl
```diff
@@ -0,0 +1,57 @@
...
+"""
+function edgeinduced_subgraphs_no_leaves(g::AbstractNamedGraph, max_number_of_edges::Int64)
+  edge_subgraphs = unique_cyclesubgraphs_limited_length(g, max_number_of_edges)
+  isempty(edge_subgraphs) && return []
```
mtfishman, line 41:
```suggestion
  isempty(edge_subgraphs) && return typeof(g)[]
```
to make it a concrete type (I assume generally it will be a collection of graphs of the same type as the input graph?).

src/simplecycles.jl
```diff
@@ -0,0 +1,57 @@
...
+
+  #Take powerset, but don't exceed max_number of edges and remove disconnected components
+  min_loop_size = minimum(length.(edges.(edge_subgraphs)))
+  max_genus = round(Int64, ceil(max_number_of_edges / min_loop_size))
+  subgraph_components = collect(powerset(edge_subgraphs, 1, max_genus))
+
+  edge_subgraphs = []
```
mtfishman, line 48:
```suggestion
  edge_subgraphs = typeof(g)[]
```

src/simplecycles.jl
JoeyT1994: Ah yeah the graph `g` is not needed. It just converts a list of vectors (that represent a cycle) to the corresponding path in terms of `NamedEdges`. I have removed the first `arg` now.

mtfishman:
Looks good!

PR #98, opened 2025-04-25
Edge coloring

JoeyT1994:
This PR wraps the edge_color algorithm implemented in SimpleGraphAlgorithms.jl (see https://github.com/scheinerman/SimpleGraphAlgorithms.jl) to work for the AbstractGraph type (including SimpleGraphs and AbstractNamedGraphs).

Testing is added and the code is added in an external module so as to avoid having SimpleGraphAlgorithms.jl and SimpleGraphConverter.jl as direct dependencies. This makes opens up the possibility of utilizing further algorithms in SimpleGraphAlgorithms.jl.


ext/NamedGraphsSimpleGraphAlgorithmsExt/NamedGraphsSimpleGraphAlgorithmsExt.jl
```diff
@@ -0,0 +1,17 @@
+module NamedGraphsSimpleGraphAlgorithmsExt
+using Graphs: AbstractGraph, edgetype
+using NamedGraphs: position_graph, vertices
+using NamedGraphs.GraphsExtensions: GraphsExtensions
+using SimpleGraphAlgorithms: SimpleGraphAlgorithms
+using SimpleGraphConverter: UG
+
+function SimpleGraphAlgorithms.edge_color(g::AbstractGraph, k::Int64)
```
mtfishman, line 8:
```suggestion
function SimpleGraphAlgorithms.edge_color(g::AbstractNamedGraph, k::Int64)
```
Otherwise this would be [type piracy](https://docs.julialang.org/en/v1/manual/style-guide/#avoid-type-piracy).

ext/NamedGraphsSimpleGraphAlgorithmsExt/NamedGraphsSimpleGraphAlgorithmsExt.jl
```diff
@@ -0,0 +1,17 @@
+module NamedGraphsSimpleGraphAlgorithmsExt
+using Graphs: AbstractGraph, edgetype
+using NamedGraphs: position_graph, vertices
+using NamedGraphs.GraphsExtensions: GraphsExtensions
+using SimpleGraphAlgorithms: SimpleGraphAlgorithms
+using SimpleGraphConverter: UG
```
mtfishman, line 6:
```suggestion
using SimpleGraphConverter: UndirectedGraph
```
I'd prefer to use the longer name so the code is more self-explanatory.

mtfishman:
Looks great! This will be very useful.

PR #104, opened 2025-09-24
PartitionsGraphView

JoeyT1994:
This PR adds a `PartitionedGraphView` wrapper for the `PartitionsGraph` such that the higher level graph is the one now exposed to the standard `Graphs.jl` functions such as `vertices, edges, neighbors ...` etc..

A few tests are added too.

@mtfishman Let me know how much functionality you think should be ported over to this wrapper, or if we keep it minimal and on an "as needed" basis as we try to use this in `ITensorNetworksnext`


Project.toml
```diff
@@ -1,14 +1,17 @@
...
 
 [deps]
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
 Combinatorics = "861a8166-3701-5b0c-9a16-15d98fcdc6aa"
 Dictionaries = "85a47980-9c8c-11e8-2b9f-f7ca1fa99fb4"
 Graphs = "86223c79-3864-5bf0-83f7-82e725a168b6"
+JuliaFormatter = "98e50ef6-434e-11e9-1051-2b60c6c9e899"
+KaHyPar = "2a6221f6-aa48-11e9-3542-2d9e0ef01880"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+Metis = "2679e427-3c69-5b7f-982b-ece356f1e94b"
```
mtfishman, line 14: These look like mistakes.

src/lib/PartitionedGraphs/src/partitionedgraphview.jl
```diff
@@ -0,0 +1,36 @@
+struct PartitionedGraphView{V,PG<:AbstractGraph{V}} <: AbstractGraph{V}
+  value::PG
+end
+
+value(pgv::PartitionedGraphView) = pgv.value
+Base.copy(pgv::PartitionedGraphView) = PartitionedGraphView(copy(value(pgv)))
+PartitionedGraph(pgv::PartitionedGraphView) = value(pgv)
```
mtfishman, line 7:
```suggestion
  partitioned_graph::PG
end

PartitionedGraph(pgv::PartitionedGraphView) = pgv.partitioned_graph
Base.copy(pgv::PartitionedGraphView) = PartitionedGraphView(copy(PartitionedGraph(pgv)))
```

src/lib/PartitionedGraphs/src/partitionedgraphview.jl
```diff
@@ -0,0 +1,35 @@
...
+end
+
+#Functionality that behaves the same
+for f in [:unpartitioned_graph]
+  @eval begin
+    function $f(pgv::PartitionedGraphView, args...; kwargs...)
+      return $f(PartitionedGraph(pgv), args...; kwargs...)
+    end
+  end
+end
```
mtfishman, line 35: I don't think this is the right definition of `underlying_graph` since a `PartitionsGraphView` should just act like the graph of partitions, so the underlying graph is just itself.

PR #147, opened 2026-03-13
New interface for `GraphsExtensions.similar_graph`; misc bug fixes.

jack-dunham:
qThis PR overhauls the way the `GraphsExtensions.similar_graph` function behaves, and includes some minor bug fixes and refactors.

##  New `similar_graph` interface

Essentially, the base case is now the three argument method, where one provides both `vertices` and `edges`. This defaults to constructing a concrete `NamedGraph` or `NamedDiGraph` depending on the `IsDirected` trait of the input graph. 
```julia
similar_graph(::AbstractGraph::!(IsDirected), vertices, edges) # -> NamedGraph
similar_graph(::AbstractGraph::IsDirected, vertices, edges) # -> NamedDiGraph
```
If edges are not provided, then the function attempts to construct a similar graph to the input graph, but with the provided vertices and no edges. If no vertices or edges are provided, then the output graph will be a similar graph with the same vertices and edges. For a given subtype of `AbstractNamedGraph`, one should overload the three argument method only (if a different return type is desired).  

##  Type domain

If instead a graph type is provided as the first argument to `similar_graph`, then the constructor is called by default:
```Julia
similar_graph(T::Type{<:AbstractGraph})  = T()
similar_graph(T::Type{<:AbstractGraph}, vertices)  = T(vertices)
similar_graph(T::Type{<:AbstractGraph}, vertices, edges)  = add_edges!(similar_graph(T, vertices), edges)
```
Note, _unlike_ the case where a graph value is provided, the one argument method is not defined in terms of the two or three argument method. This is to accommodate cases where a given `AbstractGraph` may have fixed edges, say (a notable example would be a tree graph). In that case, it is assumes there exists a constructor `T(vertices)`. As the method
```Julia
similar_graph(T::Type{<:AbstractGraph}, vertices)  = T(vertices)
```
may or may not return an edgeless graph, depending on the definition of the constructor. 
In light of this, there now exists the function
```Julia
edgeless_graph(T::Type{<:AbstractGraph}, vertices)
```
that attempts to explicitly a construct an edgeless graph, thus throwing an error when this is not possible. Likewise, there exists a function
```Julia
empty_graph(T::Type{<:AbstractGraph})
```
for explicitly constructing a graph of type `T` with no vertices or edges.

If your graph type `MyGraph` has freedom over vertices and edges, all three type-based method can be overloaded simultaneously using the followings signature:
```Julia
similar_graph(T::Type{<:MyGraph}, vertices = [], edges = [])  = ...
```


src/abstractnamedgraph.jl
```diff
@@ -1,5 +1,6 @@
 using .GraphsExtensions: GraphsExtensions, directed_graph, incident_edges,
-    partition_vertices, rem_edges!, rename_vertices, similar_graph, subgraph
+    partition_vertices, rem_edges!, rename_vertices, similar_edgeless_graph, similar_graph,
```
mtfishman:
```suggestion
    partition_vertices, rem_edges!, rename_vertices, similar_graph,
```
Looks like a stale import.

src/abstractnamedgraph.jl
```diff
@@ -44,28 +45,63 @@ ordered_vertices(graph::AbstractSimpleGraph) = vertices(graph)
...
+    V = eltype(vertices)
+    graph = add_edges!(NamedDiGraph{V}(vertices), edges)
+    return graph
+end
+
+# Passing a type as a first argument attempts to call a constructor. Should be overloaded
+# if the constructor doesnt exist for a given `AbstractGraph` concrete type.
+similar_graph(T::Type{<:AbstractGraph}) = T()
+similar_graph(T::Type{<:AbstractGraph}, vertices) = T(vertices)
+similar_graph(T::Type{<:AbstractGraph}, vertices, edges) = T(vertices, edges)
```
mtfishman: I started writing a comment about this but can't remember if I sent it already so apologies if this is a repeat.

I think there's a slightly tricky question here about what kinds of constructors and operations we should assume of a given graph type.

`T()` and `T(vertices)` seems reasonable, however `T(vertices, edges)` seems a bit strange to me (do our graph types such as NamedGraph even have that constructor?).

My feeling would be that a better set of definitions would be something like this:
```julia
similar_graph(T::Type{<:AbstractGraph}) = T()
similar_graph(T::Type{<:AbstractGraph}, vertices) = T(vertices)
function similar_graph(T::Type{<:AbstractGraph}, vertices, edges)
    g = similar_graph(T, vertices)
    add_edges!(g, edges)
    return g
end
```
though definitely let me know your thoughts on that, I don't think the choice is obvious. The only reference point I know of for this design is how `similar` handles array type inputs but I think it is pretty ad-hoc (there it assumes an array has a constructor `MyArray(undef, axes)` as a fallback).

src/abstractnamedgraph.jl
```diff
@@ -44,28 +45,63 @@ ordered_vertices(graph::AbstractSimpleGraph) = vertices(graph)
...
+edgeless_graph(graph::AbstractGraph) = rem_edges!(copy(graph), edges(graph))
+# The intention is this will fail if `T` cannot be edgeless.
+edgeless_graph(T::Type{<:AbstractGraph}, vertices) = similar_graph(T, vertices, [])
+
+empty_graph(graph::AbstractGraph) = rem_vertices!(copy(graph), vertices(graph))
+# The intention is this will fail if `T` cannot be empty.
+empty_graph(T::Type{<:AbstractGraph}) = similar_graph(T, vertextype(T)[], [])
+
+function Base.copy(graph::AbstractNamedGraph)
+    return similar_graph(typeof(graph), vertices(graph), edges(graph))
```
mtfishman: Couldn't this just be `similar_graph(graph)`?

src/abstractnamedgraph.jl
jack-dunham: I was unsure about this too, so went with the "dumber" option w.r.t to the type domain `similar_graph` methods. Although, my thought process on this is that  `T(vertices, edges)` may be the only sensible way to construct a given `T`, if for example there was a restriction on the which edges are allowed given a set of vertices, so it seems a bit arbitrary to treat `T(vertices)` as "more important" (for lack of better terminology). I could certainly see the case for making:
```Julia
function similar_graph(T::Type{<:AbstractSubTypeGraph}, vertices, edges)
    g = similar_graph(T, vertices)
    add_edges!(g, edges)
    return g
end
```
for any given `AbstractSubTypeGraph <: AbstractGraph`, (an example is `AbstractSimpleGraph` which *does* have this behavior (via `similar_simplegraph`).

src/abstractnamedgraph.jl
jack-dunham: No, as we went a graph of the same type as `graph`, so we call the type domain `similar_graph`, which needs the vertices and edges passed to it. The method `similar_graph` might construct a `NamedGraph` if one has not specialized on `similar_graph` for a given graph type, which would  then not be a copy of the given graph.

jack-dunham: This function was originally defined as `not_implemented()` but there was a todo comment to implement it generically in terms of what has no become `similar_graph`, but now I am wondering if it is best to leave it unimplemented instead of defining a fallback, as this fallback will be wrong for e.g. `AbstractDataGraph` (we could defined `Base.copy(::AbstractDataGraph) = not_implemented()` in that case, however the issue remains if one defines some `AbstractGraph` subtype  with mutable data that is uninitialized by default). 

If you agree I'll revert this commit and instead remove the todo comment.

src/abstractnamedgraph.jl
mtfishman: Right, a good example for where `add_edges!(g, edges)` would fail would be an immutable graph (like a grid graph type). But for those cases I think the right thing to do would be to overload `similar_graph(T::Type{<:AbstractGraph}, vertices, edges)`. So I think one way to think about these fallback `similar_graph` definitions is, what functionality can we expect _most_ (named) graph types to define? `T(vertices, edges)` is kind of arbitrary even for immutable graphs since a grid graph type wouldn't define that either.

mtfishman: Basically, these are fallback definitions, so we want them to work in the average case so most standard graph types don't have to define them, so the goal isn't to have them work in all cases, just most cases.

jack-dunham: I think that is a compelling argument.

src/abstractnamedgraph.jl
mtfishman: Those are good points to consider.

Regarding the type issue, `Base.copy(a::AbstractArray)` falls back to `copyto!(similar(a), a)` so I think we can take an analogous approach and not worry too much about the type (types can always overload `copy` if `similar_graph` is not strict enough, we're just looking for generic definitions right now). Also, even though the type version of `similar_graph` is stricter as a fallback, I don't know if it will be in general.

Given that `Base.copy(a::AbstractArray)` falls back to `copyto!(similar(a), a)` we probably we want the graph analog of that, do we have a version of `copyto!` for graph data? The "safe" option could be something like:
```julia
Base.copy(graph::AbstractNamedGraph) = copyto!(similar_graph(graph), graph)

# Special cases with no graph data:
Base.copy(graph::NamedGraph) = similar_graph(graph)
```

mtfishman: Sorry, I missed part of your message. It's an interesting point that `similar_graph` might fall back to `NamedGraph` and therefore drop data. That probably points even more to a fallback definition `Base.copy(graph::AbstractNamedGraph) = copyto!(similar_graph(graph), graph)`, since then the `copyto!` would fail and it would tell whoever made that graph type that they need to overload `copy`, `similar_graph`, and/or `copyto!`.

mtfishman: If we don't want to work on `copyto!` or something similar in this PR we could keep `Base.copy(graph::AbstractNamedGraph) = not_implemented()` and work out a design like that in a future PR, but I would like to work out a plan so that we have an idea for what to do and don't make a choice that we'll need to backtrack on.

jack-dunham: The `copyto!` idea makes sense to me, thanks for the input. As we also have a `DataGraphs` PR open, I think it is sensible to just do this now.

mtfishman: Makes sense. I guess `copyto!` isn't much of a stretch to add since it will fail for NamedGraphs and for DataGraphs the definition is related to broadcasting operations you already worked on.

mtfishman:
(I closed and reopened the PR to try to trigger the CI again, I'm trying to fix the docs build.)

mtfishman:
Looks good, thanks!

PR #169, opened 2026-05-14
Add functions `subgraph_edges` and `forest_cover_edge_sequence`; refine `similar_graph` interface.

jack-dunham:
This PR does the following:
1. removes the `similar_graph` methods that took edges as an argument (these were removed in DataGraphs.jl , but not in NamedGraphs.jl yet.), 
2. adds the `subgraph_edges` function.
3. adds the `forest_cover_edge_sequence` function, that is currently sitting out-of-place in ITensorNetworksNext.jl, into NamedGraphs.GraphsExtensions.

### DIFF
--- a/src/abstractnamedgraph.jl
+++ b/src/abstractnamedgraph.jl
@@ -1,4 +1,4 @@
-using .GraphsExtensions: GraphsExtensions, directed_graph, incident_edges,
+using .GraphsExtensions: GraphsExtensions, all_edges, directed_graph, incident_edges,
     partition_vertices, rem_edges, rem_edges!, rem_vertices, rename_vertices,
     similar_simplegraph, subgraph
 using Dictionaries: set!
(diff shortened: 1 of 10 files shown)


src/abstractnamedgraph.jl
```diff
@@ -52,56 +52,39 @@ GraphsExtensions.convert_vertextype(::Type{V}, g::AbstractNamedGraph{V}) where {
 GraphsExtensions.convert_vertextype(::Type, g::AbstractNamedGraph) = not_implemented()
 
 function similar_graph(graph::AbstractGraph)
-    return similar_graph(graph, copy(vertices(graph)), copy(edges(graph)))
-end
-
-function similar_graph(graph::AbstractGraph, vertices)
-    new_edge_type = convert_vertextype(eltype(vertices), edgetype(graph))
-    return similar_graph(graph, vertices, new_edge_type[])
+    newgraph = similar_graph(graph, copy(vertices(graph)))
```
mtfishman: Is this the right place for this call to `copy`? Maybe that should be the job of `similar_graph(graph, vertices)` or the constructor `similar_graph(graph, vertices)` calls out to? I actually hadn't thought much about how mutation and data sharing of vertices should be handled in general...

mtfishman: My guess would be that the "proper" place to have this copy is in the constructor itself, i.e. `NamedGraph(vertices)` should wrap a copy rather than a view of `vertices`.

src/abstractnamedgraph.jl
```diff
@@ -52,56 +52,39 @@ GraphsExtensions.convert_vertextype(::Type{V}, g::AbstractNamedGraph{V}) where {
...
-    return graph
-end
+similar_graph(T::Type{<:AbstractGraph}, vertices = vertextype(T)[]) = T(vertices)
 
 # If `T <: AbstractSimpleGraph`, then we defer to `GraphsExtensions.similar_simplegraph`.
-function similar_graph(T::Type{<:AbstractSimpleGraph}, vertices = 0, edges = [])
-    return similar_simplegraph(T, vertices, edges)
+function similar_graph(T::Type{<:AbstractSimpleGraph}, vertices = 0)
+    return similar_simplegraph(T, vertices)
 end
```
mtfishman: I think for clarity these should be split into separate functions, i.e. `similar_graph(T::Type{<:AbstractGraph}) = similar_graph(T, vertextype(T)[])` and `similar_graph(T::Type{<:AbstractGraph}, vertices) = T(vertices)`.

src/abstractnamedgraph.jl
jack-dunham: Hmm I'm not sure either. My thoughts was that the new graph obtained via `similar_graph` should definitely by independent of the input graph (hence the copy only when the vertices are obtained directly from the input graph), whereas if one is just providing a list of vertices then you should kind of expect that maybe mutating that vertex list might alter the graph (as the graph has been constructed from it, after all). 
```Julia
similar_graph(graph) # construct a new graph like `graph`
similar_graph(graph, vertices) # construct a new graph like `graph` but with `vertices`.
```
I think it makes sense that the first method should copy the list of vertices from `graph`, but the second method should not necessarily have to copy the vertices input as the argument.

src/abstractnamedgraph.jl
```diff
@@ -413,12 +396,12 @@ function Graphs.has_path(
...
 
 function Base.union(graph1::AbstractNamedGraph, graph2::AbstractNamedGraph)
-    union_graph = promote_type(typeof(graph1), typeof(graph2))()
+    union_graph_type = promote_type(typeof(graph1), typeof(graph2))
```
mtfishman: For `Base.union`, it bases the output type on the first argument:
```julia
julia> union(Set([1, 2, 3]), [1.0, 2.0, 3.0])
Set{Float64} with 3 elements:
  2.0
  3.0
  1.0

julia> union([1, 2, 3], Set([1, 2, 3]))
3-element Vector{Int64}:
 1
 2
 3
```
Maybe we could use that here? `promote_type` seems a bit fickle to use here. I.e. we could just change it to `union_graph = similar_graph(graph1, union_vertices)`.

src/abstractnamedgraph.jl
jack-dunham: > My guess would be that the "proper" place to have this copy is in the constructor itself, i.e. `NamedGraph(vertices)` should wrap a copy rather than a view of `vertices`.

I think that is an opinionated decision that a given concrete graph could make, but I think in the case of `similar_graph(graph)`, the vertices should *always* be copied, hence the call to `copy`.

mtfishman: I think the question comes down to what we should expect of graph types, and then where the customization points are. I think it is pretty reasonable to expect that for a graph type `G <: AbstractNamedGraph`, `G(vertices)` copies the vertices (it's a pretty standard assumption for constructors to copy inputs unless the type is very clearly a wrapper, but I wouldn't consider `NamedGraph(vertices)` to be a wrapper around the vertice). Then, if a graph type differs from that, I think the right customization point for doing the copy would be `similar_graph(graph, vertices)`. Otherwise, how do we decide which generic functions should call `copy(vertices(graph))` before passing those vertices to a graph constructor? We'd have to do that everywhere.

jack-dunham: > Otherwise, how do we decide which generic functions should call `copy(vertices(graph))` before passing those vertices to a graph constructor? We'd have to do that everywhere.

That's a good point. I'll update the PR with this in mind.

PR #170, opened 2026-05-20
Fix infinite method call in `forest_cover_edge_sequence`; add tests.

jack-dunham:
The `forest_cover_edge_sequence` method dispatching on `AbstractNamedGraph` would call itself repeatedly as there was no `forest_cover_edge_sequence` method dispatching on concrete `NamedGraph`. 

I've removed this method as it was also not necessarily given the generic implementation of `forest_cover_edge_sequence` on `AbstractGraph`, and added some simple tests for this function. 

This PR also refactors the implementation of `similar_graph`, moving this function back into `GraphsExtensions`.


mtfishman:
Looks good, ready to merge? Would this mean we don't need to define `forest_cover_edge_sequence` for AbstractDataGraph, i.e. in https://github.com/ITensor/DataGraphs.jl/pull/121?

src/lib/GraphsExtensions/src/abstractgraph.jl
```diff
@@ -55,16 +64,60 @@ end
...
+# a similar graph type that has no notion of data (in the abstract sense).
+
+function similar_dataless_graph(graph::AbstractGraph)
+    dataless_graph = similar_graph(graph, vertices(graph))
+    add_edges!(dataless_graph, edges(graph))
+    return dataless_graph
+end
+function similar_dataless_graph(graph::AbstractGraph, vertices)
+    return throw(MethodError(similar_dataless_graph, (typeof(graph), typeof(vertices))))
+end
```
mtfishman, line 87: I know we discussed this but I can't remember the design we had decided on... This looks slightly off to me, I think this would make more sense to me:
```julia
function similar_dataless_graph(graph::AbstractGraph)
    dataless_graph = similar_dataless_graph(graph, vertices(graph))
    add_edges!(dataless_graph, edges(graph))
    return dataless_graph
end
function similar_dataless_graph(graph::AbstractGraph, vertices)
    return similar_graph(graph, vertices)
end
```
So then both "just work" in general, and `similar_dataless_graph(graph::AbstractGraph, vertices)` is the minimal overload for graphs that need to customize.

jack-dunham: My idea is that unless `vertices` is an `Int` or a `Base.OneTo` (which have methods defined), we don't really know how to construct a good similar dataless graph (within the context of `GraphsExtensions`), so this function should error. 

However, forwarding to `similar_graph` in this case is a good idea (which also errors for general vertices, but can be overloaded not to for a given `AbstractGraph` subtype.

mtfishman:
Looks like some downstream tests are failing, is the goal to make this PR non-breaking?

src/similar_graph.jl
```diff
@@ -0,0 +1,70 @@
...
+    )
+    return NamedGraph(vertices)
+end
+
+@traitfn function GraphsExtensions.similar_dataless_graph(
+        graph::AbstractNamedGraph::IsDirected,
+        vertices
+    )
+    return NamedDiGraph(vertices)
+end
```
mtfishman: I think we could get rid of these definitions if we use the design suggested here: https://github.com/ITensor/NamedGraphs.jl/pull/170/changes#r3311245873

src/lib/GraphsExtensions/src/abstractgraph.jl
```diff
@@ -55,16 +64,60 @@ end
...
+@traitfn function directed_graph(graph::AbstractGraph::(!IsDirected))
+    digraph = similar_graph(directed_graph_type(graph), vertices(graph))
+    add_edges!(digraph, all_edges(graph))
+    return digraph
+end
+@traitfn function undirected_graph(graph::AbstractGraph::IsDirected)
+    undigraph = similar_graph(undirected_graph_type(graph), vertices(graph))
+    for e in edges(graph)
+        has_edge(undigraph, e) && continue
+        add_edge!(undigraph, e)
```
mtfishman, line 117: Does `add_edge!` error if the edge exists already?

mtfishman: I.e. can't this be `add_edges!(digraph, edges(graph))`?

src/similar_graph.jl
jack-dunham: We can, but it then means that if we change `similar_graph` then, the behavior of `similar_dataless_graph` changes. I try to take the perspective that a function should call another function because it makes sense to call that function, not because the code behind each function happens to be shared.

jack-dunham:
The `ITensorNetworksNext` tests failing is strange, i'll investigate. 

I am less bothered if `DataGraphs` breaks and needs updated slightly as we have an open PR for that package anyway, so can upgrade to the new NamedGraphs version (assuming that doesn't breaking downstream tests).

src/lib/GraphsExtensions/src/abstractgraph.jl
jack-dunham: It should *not* error according to `Graphs.jl`, however I was worried about inconsistency w.r.t. this so thought the above would be safer for now.

src/similar_graph.jl
jack-dunham: I guess one can argue that `similar_dataless_graph` should fallback to `similar_graph` by default. Happy to change.

mtfishman:
> The `ITensorNetworksNext` tests failing is strange, i'll investigate.
> 
> I am less bothered if `DataGraphs` breaks and needs updated slightly as we have an open PR for that package anyway, so can upgrade to the new NamedGraphs version (assuming that doesn't breaking downstream tests).

I agree it isn't a huge deal if this PR is breaking, I just wanted to check if that is deliberate, check how easy it would be to make it non-breaking, inquire what specifically about the PR is breaking, etc. But also if it does break DataGraphs we should mark this PR as breaking with a minor version bump.

src/similar_graph.jl
mtfishman: I guess I was picturing a design where we assume `AbstractNamedGraph` is dataless, and then if it has data it should define `similar_dataless_graph`. But, I could see an argument for the current design.

jack-dunham:
> inquire what specifically about the PR is breaking

For `DataGraphs`, there's a test of the function `empty_graph` on a `DataGraph` (checks this function removes data). I have moved this function from `NamedGraphs` to `GraphsExtensions`, as is generic to any `AbstractGraph`, so the import statement in the tests fails.

That should be the only place where code is breaking AFAIK. We can reverse this change, but it seems logical to move that function in this PR.

mtfishman:
> > inquire what specifically about the PR is breaking
> 
> For `DataGraphs`, there's a test of the function `empty_graph` on a `DataGraph` (checks this function removes data). I have moved this function from `NamedGraphs` to `GraphsExtensions`, as is generic to any `AbstractGraph`, so the import statement in the tests fails.
> 
> That should be the only place where code is breaking AFAIK. We can reverse this change, but it seems logical to move that function in this PR.

Can we add a statement `using .GraphsExtensions: empty_graph` in `NamedGraphs` for backwards compatibility?

src/lib/PartitionedGraphs/src/quotientvertex.jl
```diff
@@ -79,7 +79,7 @@ function Graphs.vertices(g::AbstractGraph, quotientvertex::QuotientVertex)
     qv = parent(quotientvertex)
 
     pvs = partitioned_vertices(g)
-    haskey(pvs, qv) || throw(ArgumentError("Quotient vertex $quotientvertex not in graph"))
+    qv ∈ keys(pvs) || throw(ArgumentError("Quotient vertex $quotientvertex not in graph"))
```
mtfishman, line 84: Could you make a comment in the code about this change so we remember why it was needed?

src/lib/PartitionedGraphs/src/quotientvertex.jl
```diff
@@ -88,7 +88,7 @@ function Graphs.vertices(g::AbstractGraph, quotientvertices::QuotientVertices)
 end
 
 function has_quotientvertex(g::AbstractGraph, quotientvertex::QuotientVertex)
-    return haskey(partitioned_vertices(g), parent(quotientvertex))
+    return parent(quotientvertex) ∈ keys(partitioned_vertices(g))
```
mtfishman, line 94: Same thing here.

src/lib/GraphsExtensions/src/abstractgraph.jl
```diff
@@ -55,16 +64,61 @@ end
...
+    return similar_graph(T, length(vertices))
+end
+similar_graph(T::Type{<:AbstractSimpleGraph}) = similar_graph(T, 0)
+
+# =============================== `similar_dataless_grap` ================================ #
+# This function behaves much the same as `similar_graph`, but should strictly return a
+# a similar graph type that has no notion of data (in the abstract sense).
+
+function similar_dataless_graph(graph::AbstractGraph)
+    dataless_graph = similar_graph(graph, vertices(graph))
```
mtfishman:
```suggestion
    dataless_graph = similar_dataless_graph(graph, vertices(graph))
```
?

src/lib/GraphsExtensions/src/trees_and_forests.jl
```diff
@@ -47,9 +47,12 @@ end
 # Given an undirected graph g with vertex set V, build a set of forests (each with vertex set V) which covers all edges in g
 # (see https://en.wikipedia.org/wiki/Arboricity) We do not find the minimum but our tests show this algorithm performs well
 function forest_cover(g::AbstractGraph; spanning_tree = spanning_tree)
-    edges_collected = edgetype(g)[]
+    g = similar_dataless_graph(g)
+    g_reduced = copy(g)
```
mtfishman: Probably doesn't matter too much, but I don't see what this copy is needed for since it doesn't look like the graph is being modified in-place. If it isn't needed we can just write `g_reduced = g` if the point is primarily to have a different variable name to use. I would just find that clearer since when I saw a copy I then expected to see an in-place operation somewhere.

jack-dunham: Good point.

Issue #175, opened 2026-07-07
[BUG] Edge_subgraph fails due to implicitly caring about edge directionality on undirected graphs


JoeyT1994:
In the current version of the library (0.12), `edge_subgraph(g::NamedGraph, es::AbstractVector)` is returning incorrect results.

For instance the last assertion here fails. 

```
using NamedGraphs
using NamedGraphs.NamedGraphGenerators: named_hexagonal_lattice_graph
using NamedGraphs: NamedEdge, edges
using NamedGraphs.GraphsExtensions: edge_subgraph

g = named_hexagonal_lattice_graph(3, 3)
#Ring of six edges in the graph
es = NamedEdge.([(1, 1) => (2, 1), (2, 1) => (3, 1), (3, 1) => (3, 2), (2, 2) => (3, 2), (1, 2) => (2, 2), (1, 1) => (1, 2)])

#They are all in `g` (and specified in the correct direction)
@assert all([e ∈ edges(g) for e in es])

#Build the subgraph from these edges
eg = edge_subgraph(g,es)

#But now one of them is missing from the subgraph they induce...
@assert length(edges(eg)) == length(es)
```
but passes on older versions of the library (0.7 for instance).

This appears to be due to the fact that internally in `edge_subgraph` the call to  `_g = subgraph(g, vs)` builds a subgraph with edges that may be defined in reverse compared to the original graph `g` and then the edge subgraph is built by removing a `setdiff(edges(g), es)` from `_g` which cares about edge directionality.

PR #176, opened 2026-07-08
Fix edge subgraph

JoeyT1994:
This PR fixes Issue #175 by defining `edge_subgraph(g, edgelist)` separately for the `NamedGraph` and `NamedDiGraph` types, which should behave differently because one cares about the direction of edges in the latter but not in the former.


src/abstractnamedgraph.jl
```diff
@@ -573,6 +573,10 @@ end
 function edge_subgraph_namedgraph(graph, edgelist)
     vs = unique(vcat(src.(edgelist), dst.(edgelist)))
     g = subgraph(graph, vs)
-    g = rem_edges!(g, setdiff(edges(g), edgelist))
+    for e in edges(g)
+        if !(e ∈ edgelist || reverse(e) ∈ edgelist)
```
mtfishman: My concern here would be efficiency, i.e. the previous version used set operations which use hashing, which makes lookup faster. Maybe a simple tweak could be to define `edgeset = Set(edgelist)` and then use `e ∈ edgeset`, etc.

JoeyT1994:
@mtfishman This is good to merge on my end. I don't understand what the downstream ITensorNetworks test fail is, seems like something to do with `OMEinsum` so not sure its related to this PR?

mtfishman:
> @mtfishman This is good to merge on my end. I don't understand what the downstream ITensorNetworks test fail is, seems like something to do with `OMEinsum` so not sure its related to this PR?

Sorry for the delay, this is good to go from my end but I wanted to look into that before merging, not sure what's going on there. (Looks like it was a temporary glitch using some incompatible package versions.)

PR #177, opened 2026-07-13
Faster loop enumeration

JoeyT1994:
This PR replaces the `edgeinduced_subgraphs_no_leaves(g::AbstractGraph, max_edges::Integer)` function with a much faster and fully correct implementation.

The function takes a graph `g` and enumerates all connected, edge induced subgraphs with up to `max_edges` which contain no leaf (degree 1) vertices. This induced graphs are the basis of a cluster expansion.

The function is now completely correct (the previous version missed out subgraphs with bridges) and much much faster (many hundreds of times on large graphs).

Tests are updated to reflect this and check the expansion.


src/simplecycles.jl
```diff
@@ -1,59 +1,192 @@
...
+    const vertex_degree::Vector{Int}
+    # Number of vertices whose induced degree is exactly 1 (i.e. leaves of the current set).
+    num_leaves::Int
+    # `marked[u]` is true while edge `u` is in the current set or already queued for it.
+    const marked::Vector{Bool}
+    # The current edge set, used as a stack.
+    const current_edges::Vector{Int}
+    # Completed leaf-free edge sets (each a vector of edge indices).
+    const results::Vector{Vector{Int}}
+end
```
mtfishman: Since most of the fields are set to `const` anyway, it might be better to make the type not mutable and then make the field `num_leaves::Base.RefValue{Int}`. Then you can update it with `search.num_leaves[] += 1` and access it with `search.num_leaves[]`.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "NamedGraphs"
 uuid = "678767b0-92e7-4007-89e4-4527a8725b19"
-version = "0.12.2"
+version = "0.12.3"
```
mtfishman: Since we are deleting `unique_simplecycles_limited_length` I think we'll need to mark this as breaking, i.e. `v0.13`. I guess the idea is that function is just not very useful so we may as well not support it any more?

mtfishman:
@JoeyT1994 what do you think of the name `leafless_edge_induced_subgraphs`?

Project.toml
JoeyT1994: Yeah it was only being used inside the previous version of `edgeinduced_subgraphs_no_leaves` so seems no point in supporting it. Will bump it

JoeyT1994:
@mtfishman Yeah that's a slightly simpler name, I'll change to that

